PR #1, opened 2019-04-02
Start scaffolding repo structure

morgante:


ocsig:
Do we want templates/dm, templates/tf and not dm/templates, tf/templates?

morgante:
I'm fine with either.

PR #4, opened 2019-04-09
Resolved issues with instance schema naming not aligned with api

t04glovern:
Schema naming for `natIP` and `networkIP` were incorrectly named `natIp` and `networkIp`.

Not an issue with the actual API requests, however having a matching schema makes it a lot easier to troubleshoot the problem.


ocsig:
Great catch, would you mind to update the instance template and managed instance group templates+schemas?

morgante:
@ocsig @t04glovern Ping

t04glovern:
I could only find a reference in the managed instance template. I've updated accordingly :)

PR #33, opened 2019-05-16
Initial implementation of CFT cli tool on golang 

kopachevsky:
Ready for review


cli/go.sum
```diff
@@ -0,0 +1,5 @@
+github.com/kopachevsky/cloud-foundation-toolkit v0.0.0-20190515212103-a60e275a7fa7 h1:EmKqOddOW2st/vtj3RgDoaJbll7YwkS46MIVghaejKk=
```
kopachevsky: just noticed, selflink should not be here

cli/Makefile
```diff
@@ -0,0 +1,12 @@
+SHELL   := /bin/bash
+
+# need to change this to VERSION=`git describe --tags`
```
morgante: Hard-coded version is probably fine as well, as long as we are clear on how it can be updated.

cli/cft.go
```diff
@@ -0,0 +1,12 @@
+package main
```
morgante: Should this file be `main.go` by convention?

cli/cmd/root.go
```diff
@@ -0,0 +1,22 @@
...
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+// rootCmd represents the base command when called without any subcommands
+var rootCmd = &cobra.Command{
+	Use:   "cft",
+	Short: "cft",
```
morgante: As per [examples](https://github.com/spf13/cobra/blob/master/cobra/cmd/init.go#L29), short should be a short description (probably line 14 works) while `Long` is a full explanation of functionality.

cli/go.sum
morgante: Yeah, let's remove this or figure out why it got added.

kopachevsky: removed

cli/Makefile
kopachevsky: agreed, compile time version set as `git describe --tags` important when you develop library that should be available with `go get` from github.com, in this case it's more easy to keep versioning only in one place: git tag, rather than in both tags and source code, lets leave it as it is now than and decide later?

cli/cft.go
kopachevsky: it's always followed by conversion, for example no much main.go files in kubernetes repos

changed to main.go for more clarify

cli/cmd/root.go
kopachevsky: updated

kopachevsky:
Ready for review! @ocsig @morgante  please take a look

cli/Makefile
```diff
@@ -0,0 +1,16 @@
+SHELL := /bin/bash
+
+# need to change this to VERSION=`git describe --tags`
+VERSION=v1.0.0
```
morgante: Let's start with v0.1.0.

cli/cmd/root.go
```diff
@@ -0,0 +1,55 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		// no params means same as -h flag
+		if args == nil || len(args) == 0 {
+			cmd.HelpFunc()(cmd, args)
+		}
+	},
+}
+
+func init() {
+	rootCmd.SetUsageTemplate(`Usage:{{if .Runnable}}{{.UseLine}}{{end}}
```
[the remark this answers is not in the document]
kopachevsky: @morgante Adam was asked to change default output of usage from **command [command]** to **command [command] [flags]**

cli/Makefile
kopachevsky: done

PR #42, opened 2019-05-22
Disallowed unexpected properties in json schema

alexkonkin:


ocsig:
Are the integration tests still passing after these changes?

alexkonkin:
Adam,
I have checked this feature with:
instance module
and
two scenarios, 
one positive (should pass) and one negative (should fail).

For the positive test I simply executed the test as is + the feature that had been added.
The result of the test was as it had been expected initially: the test passed

For the negative test I introduced a new property (dummy: demo).
The result of the test was as had been expected: the test failed,
please see the screenshot below:
https://www.screencast.com/t/uVgKwEpeSc

PR #72, opened 2019-05-27
[WIP] DM/cli Cross-deployment referencing

kopachevsky:
### Tasks in progress:

Cross-deployment referencing #story
Receive project name as input from CFT DM yaml  #story
Input multiple DM templates at once #story


cli/Makefile
```diff
@@ -4,6 +4,7 @@ SHELL := /bin/bash
 VERSION=v0.1.0
 BINARY=bin/cft
 GITHUB_REPO=github.com/GoogleCloudPlatform/cloud-foundation-toolkit
+GOOGLE_APPLICATION_CREDENTIALS=
```
ocsig: Is this an SA key? can't we get the credentials via gcloud auth login?

kopachevsky: removed to do not confuse users

PR #75, opened 2019-05-27
Update firewall.py to accept resource name being set

umairidris:
It isn't desirable to always have a hard coded resource name.


ocsig:
Let's hold this back, our team is currently updating the templates including accepting names for all tempaltes.

PR #87, opened 2019-05-31
Support requesterPays for GCS buckets

umairidris:


dm/templates/gcs_bucket/gcs_bucket.py.schema
```diff
@@ -123,6 +123,14 @@ properties:
               group|serviceAccount:email or domain:domain.
               Can also be one of the following special values: allUsers,
               allAuthenticatedUsers.
+  billing:
```
morgante: Instead of requiring the full billing object, could we collapse this with a top-level `requesterPays` boolean?

umairidris: It seems like billing currently only has requesterPays: https://cloud.google.com/storage/docs/json_api/v1/buckets/insert

Do you know if other fields are possible or in scope? Happy to pull it to top level, but wanted to keep it flexible in case the bucket billing decides to add more fields.

Just an FYI but this is already merged in the old repo before I knew about the new one: https://github.com/GoogleCloudPlatform/deploymentmanager-samples/commit/b26f0c9f8326c6b39627ed6e26a4ca735329f88c (would be nice to have some automated forwarding or something or just shutting down the old repo/putting a deprecated notice).

morgante: I don't think any other fields are possible, hence the suggestion to pull it up.

> would be nice to have some automated forwarding or something or just shutting down the old repo/putting a deprecated notice

Agreed. @ocsig What's our ETA for deprecating the old repo?

umairidris: Done.

ocsig: We are focusing on a massive code upgrade now. I wish to finish that before the merge back.
Automated mergeback is possible but not straigthforward, the two repo structure and some of the relative path are different. For now the plan is manual merge back in every few weeks.
I will communications about this soon.

PR #160, opened 2019-06-21
Added terraform_docs.sh - workaround for HCL2 & terrafrom-docs

ivankorn:
Fixes [terraform-google-module-template/issues/23](https://github.com/terraform-google-modules/terraform-google-module-template/issues/23)
- added [terraform_docs.sh](https://github.com/antonbabenko/pre-commit-terraform/blob/master/terraform_docs.sh)
- added `perl` as dependency for the script


ivankorn:
I signed it!

PR #165, opened 2019-06-25
Fix/162

bohdanyurov-gl:


dm/templates/iam_member/iam_member.py
```diff
@@ -13,28 +13,58 @@
...
 
 def generate_config(context):
     """ Entry point for the deployment resources. """
 
-    project_id = context.properties.get('projectId', context.env['project'])
+    folder_id = context.properties.get('folderId')
+    org_id = context.properties.get('organizationId')
+    project_id = context.properties.get('projectId')
+    if not folder_id and not org_id:
+        project_id = context.env['project']
```
ocsig: This means `context.properties.get('projectId')` is never used. Line 25-26 should be removed.
Line 24 should be:
`    project_id = context.properties.get('projectId', context.env['project'])`

dm/templates/iam_member/iam_member.py.schema
```diff
@@ -35,40 +39,52 @@ required:
   - roles
```
ocsig: folredID, OrgID, projectID shall not be used at the same time. Please specify this so there will be no blind overwrite.

bohdanyurov-gl: Done.

bohdanyurov-gl: This is not a problem, you can use all three if you want to. This template supports it.

dm/templates/iam_member/iam_member.py
bohdanyurov-gl: Fixed.

Issue #173, opened 2019-07-01
virtual.projects.iamMemberBinding does not work when performing a DM update


jsun6383:
### What is the problem?

When trying to update the iam policy using DM/iam_member template, deployment manager update fails with the error:

```
ERROR: (gcloud.deployment-manager.deployments.update) Error in Operation [operation-1561961486052-58c987f876cdb-81d930b3-b1a4913f]: errors:
- code: NO_METHOD_TO_UPDATE_FIELD
  message: No method found to update field 'role' on resource 'get-iam-policy-iam-member-test-0-0'
    of type 'cloudresourcemanager-v1'. The resource may need to be recreated with
    the new field.
- code: NO_METHOD_TO_UPDATE_FIELD
  message: No method found to update field 'role' on resource 'get-iam-policy-iam-member-test-1-0'
    of type 'cloudresourcemanager-v1'. The resource may need to be recreated with
    the new field.
```

### How to reproduce?

Prerequisite -
- have an existing project
- have an existing service account (with or without exist iam-binding)

create the initial binding using something like the following template: (make sure the target service account does not have the roles you are applying.

my_iammember.yaml
```
imports:
  - path: templates/iam_member/iam_member.py
    name: iam_member.py

resources:
  - name: iam-member-test
    type: iam_member.py
    properties:
      roles:
        - role: roles/owner
          members:
            - serviceAccount:<service_account_email>
        - role: roles/viewer
          members:
            - serviceAccount:<service_account_email>
        - role: roles/editor
          members:
            - serviceAccount:<service_account_email>
```

Create the deployment using gcloud:
```
gcloud deployment-manager deployments create <deployment name> --config my_iammember.yaml --project <project_id>
The fingerprint of the deployment is YDDe0jcBvNsZoSuymtHQ0A==
Waiting for create [operation-1561961867754-58c989647bc92-41977b01-1cd6e20f]...done.
Create operation operation-1561961867754-58c989647bc92-41977b01-1cd6e20f completed successfully.
NAME                                TYPE                                                                 STATE      ERRORS  INTENT
get-iam-policy-iam-member-test-0-0  gcp-types/cloudresourcemanager-v1:virtual.projects.iamMemberBinding  COMPLETED  []
get-iam-policy-iam-member-test-1-0  gcp-types/cloudresourcemanager-v1:virtual.projects.iamMemberBinding  COMPLETED  []
[...]

PR #187, opened 2019-07-03
[DM] Patch IAM member to projectID

ocsig:
Project ID was forcefully overwritten to context.env['project']

### DIFF
--- a/dm/templates/iam_member/iam_member.py
+++ b/dm/templates/iam_member/iam_member.py
@@ -24,9 +24,9 @@ def generate_config(context):
     project_id = context.properties.get('projectId', context.env['project'])
 
     resources = []
-    for ii, role in  enumerate(context.properties['roles']):
-        for i, member in enumerate(role['members']):
-            suffix =  sha1('{}-{}'.format(role['role'], member)).hexdigest()[:10]
+     for role in context.properties['roles']:
+         for member in role['members']:
+            suffix = sha1('{}-{}'.format(role['role'], member)).hexdigest()[:10]
             policy_get_name = '{}-{}'.format(context.env['name'], suffix)
 
             if org_id:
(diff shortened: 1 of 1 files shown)


ocsig:
#173 This should patch the problem:
We left the indexes in the hash generation which made a new hash for every resource when the list got changed. Indexes are no longer part of the hash, so reordering or deletion works as expected.

dm/templates/iam_member/iam_member.py
```diff
@@ -21,29 +21,26 @@ def generate_config(context):
 
     folder_id = context.properties.get('folderId')
     org_id = context.properties.get('organizationId')
-    project_id = context.properties.get('projectId')
-    if not folder_id and not org_id:
-        project_id = context.env['project']
+    project_id = context.properties.get('projectId', context.env['project'])
 
     resources = []
     for ii, role in  enumerate(context.properties['roles']):
```
sjvanrossum:
```suggestion
    for role in context.properties['roles']:
```

dm/templates/iam_member/iam_member.py
```diff
@@ -21,29 +21,26 @@ def generate_config(context):
...
         for i, member in enumerate(role['members']):
-            suffix =  sha1('{}-{}-{}-{}'.format(role['role'], member, ii, i)).hexdigest()[:10]
+            suffix =  sha1('{}-{}'.format(role['role'], member)).hexdigest()[:10]
```
sjvanrossum:
```suggestion
            suffix = sha1('{}-{}'.format(role['role'], member)).hexdigest()[:10]
```

dm/templates/iam_member/iam_member.py
```diff
@@ -21,29 +21,26 @@ def generate_config(context):
...
     folder_id = context.properties.get('folderId')
     org_id = context.properties.get('organizationId')
-    project_id = context.properties.get('projectId')
-    if not folder_id and not org_id:
-        project_id = context.env['project']
+    project_id = context.properties.get('projectId', context.env['project'])
 
     resources = []
     for ii, role in  enumerate(context.properties['roles']):
         for i, member in enumerate(role['members']):
```
sjvanrossum:
```suggestion
        for member in role['members']:
```

PR #206, opened 2019-07-09
Fix prerequisites Makefile target to align with docs and help

rosmo:
Change Makefile target from `cft-prerequisites` to `prerequisites`.

### DIFF
--- a/dm/docs/userguide.md
+++ b/dm/docs/userguide.md
@@ -370,7 +370,7 @@ Proceed as follows:
 
 ```shell
 cd dm
-sudo make prerequisites        # Installs prerequisites in system python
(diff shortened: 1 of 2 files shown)


rosmo:
It was not consistent in the docs and Makefile, but I've changed it the other way around, if that works better.

dm/docs/userguide.md
```diff
@@ -370,7 +370,7 @@ Proceed as follows:
 
 ```shell
 cd dm
-sudo make prerequisites        # Installs prerequisites in system python
+sudo make cft-prerequisites        # Installs prerequisites in system python
```
morgante: Let's keep the indenting aligned (for the comment).

rosmo: Fixed.

PR #225, opened 2019-07-10
Update docker images for network module

glarizza:
This commit updates both the lint and integration test docker images for
the network module to versions that will support the Terraform 0.12
upgrade.


glarizza:
I signed it!

Issue #233, opened 2019-07-11
The issue of "generate_*_url"


gruihuang:
We use some function called "generate_*_url" to simplify user's input. User just need to give a network name, the function will create an URI for that network. That is great when we use a single template but it brings many problems when a template work with templates.
Say we want to deploy a router that needs a network. The deployment manager must deploy the network before it create the router. The "generate_*_url" makes it impossible to use reference to specify the relationship between resources. The result is that DM has no idea if a resource should be deployed before another. 
The simplest solution is just deleting all "generate_*_url" so that users input the URL or a reference. Or we should support both "name" and "URL" so that users have a place to specify references.

gruihuang:
For example:
We might need this:
  - name: vv-test-network
    type: network.py
    properties:
      autoCreateSubnetworks: false
      subnetworks:
        - name: test-subnetwork-1
          region: us-east1
          ipCidrRange: 10.0.0.0/24
          privateIpGoogleAccess: false
          enableFlowLogs: true
  - name: test-cloud-router
    type: cloud_router.py
    properties:
      name: cloud-router
      networkURL: $(ref.vv-test-network.selfLink) #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
      region: us-central1
      asn: 65001

instead of 
  - name: vv-test-network
    type: network.py
    properties:
      autoCreateSubnetworks: false
      subnetworks:
        - name: test-subnetwork-1
          region: us-east1
          ipCidrRange: 10.0.0.0/24
          privateIpGoogleAccess: false
          enableFlowLogs: true
  - name: test-cloud-router
    type: cloud_router.py
    properties:
      name: cloud-router
      network: vv-test-network #!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
      region: us-central1
      asn: 65001

$(ref.vv-test-network.selfLink) is important for DM to figure out which one should go first.

Issue #226, opened 2019-07-10
DM VPN template supports static routing


gruihuang:
The DM VPN template supports dynamic routing only recently but we need static routing (route based and policy based) as well. Also we do not need to create a new static IP address sometimes so we wish to skip that step when an IP address is specified.

PR #234, opened 2019-07-11
Enable VPN DM template to support static routing.

gruihuang:
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/226
In this PR, we add static routing to VPN template
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/233
We also add routeURL and netURL to the schema so that users can specify the references of other resources in the config file.

### DIFF
--- a/dm/templates/vpn/vpn.py
+++ b/dm/templates/vpn/vpn.py
@@ -23,7 +23,7 @@ def generate_config(context):
 
     network = context.properties.get('networkURL', generate_network_uri(
         project_id,
-        context.properties.get('networkName','')
+        context.properties.get('network','')
     ))
     target_vpn_gateway = context.env['name'] + '-tvpng'
     esp_rule = context.env['name'] + '-esp-rule'
--- a/dm/templates/vpn/vpn.py
+++ b/dm/templates/vpn/vpn.py
@@ -32,15 +32,15 @@ def generate_config(context):
     vpn_tunnel = context.env['name'] + '-vpn'
     router_vpn_binding = context.env['name'] + '-router-vpn-binding'
     resources = []
-    static_ip = ''
-    ip_address = ''
     if 'ipAddress' in context.properties:
         ip_address = context.properties['ipAddress']
+        static_ip = ''
     else:
         static_ip = context.env['name'] + '-ip'
         resources.append({
             # The reserved address resource.
             'name': static_ip,
+            # https://cloud.google.com/compute/docs/reference/rest/v1/addresses
             'type': 'gcp-types/compute-v1:addresses',
             'properties': {
                 'region': context.properties['region']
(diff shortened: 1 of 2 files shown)


dm/templates/vpn/vpn.py
```diff
@@ -21,19 +21,34 @@ def generate_config(context):
...
     target_vpn_gateway = context.env['name'] + '-tvpng'
-    static_ip = context.env['name'] + '-ip'
     esp_rule = context.env['name'] + '-esp-rule'
     udp_500_rule = context.env['name'] + '-udp-500-rule'
     udp_4500_rule = context.env['name'] + '-udp-4500-rule'
     vpn_tunnel = context.env['name'] + '-vpn'
     router_vpn_binding = context.env['name'] + '-router-vpn-binding'
+    resources = []
+    static_ip = ''
+    ip_address = ''
```
morgante: I don't think we need to define these empty strings upfront.

dm/templates/vpn/vpn.py
```diff
@@ -21,19 +21,34 @@ def generate_config(context):
...
+    if 'ipAddress' in context.properties:
+        ip_address = context.properties['ipAddress']
+    else:
+        static_ip = context.env['name'] + '-ip'
+        resources.append({
+            # The reserved address resource.
+            'name': static_ip,
+            'type': 'gcp-types/compute-v1:addresses',
+            'properties': {
+                'region': context.properties['region']
```
morgante: Don't we still need a name and project specified?

dm/templates/vpn/vpn.py
gruihuang: ip_address can be removed.
static_ip is necessary because we need it in output.
resources is necessary because we want to use "append" and "extend" in the following codes.

morgante: Ok please remove ip_address.

dm/templates/vpn/vpn.py
gruihuang: Based on my test, the "addresses" API works with or without "project". I can add the "project" back to keep the old implementation.

morgante: If you don't specify project, it will use the DM project which isn't always desired.

ocsig:
It looks like to me you rename routerName to router in the properties, but you did not changed the code reflecting to it!

dm/templates/vpn/vpn.py.schema
```diff
@@ -81,6 +107,27 @@ properties:
     description: |
       The value is used to set the secure session between the Cloud VPN
       gateway and the peer VPN gateway.
+  localTrafficSelector:
+    type: array
+    description: |
```
nick4fake: uniqItems: True

dm/templates/vpn/vpn.py.schema
```diff
@@ -81,6 +107,27 @@ properties:
...
+      Used when establishing the VPN tunnel with the peer VPN gateway.
+    default: ["0.0.0.0/0"]
+    items:
+      type: string
+      description: "CIDR formatted string, for example: 192.168.0.0/16."
+  remoteTrafficSelector:
+    type: array
```
nick4fake: uniqItems: True

PR #235, opened 2019-07-11
Enable cloud_router DM template to support networkURL.

gruihuang:
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/233
Enable cloud_router to accept network URL.
- add a new tag "networkURL" that accepts the URL of a network.

### DIFF
--- a/dm/templates/cloud_router/cloud_router.py.schema
+++ b/dm/templates/cloud_router/cloud_router.py.schema
@@ -32,7 +32,7 @@ imports:
 additionalProperties: false
 
 required:
-  - network
+  # one of networkURL or networkName is required.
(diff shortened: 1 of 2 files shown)


dm/templates/cloud_router/cloud_router.py.schema
```diff
@@ -32,9 +32,14 @@ imports:
 additionalProperties: false
 
 required:
-  - network
   - region
 
+oneOf:
```
morgante: Since this is a breaking change, we should increment the version.

gruihuang:
Done. Please take a review.

Issue #177, opened 2019-07-02
The default GKE initialClusterVersion


gruihuang:
The schema of the GKE DM template has a default initialClusterVersion value 1.9.7-gke.6.
Is it possible that we use "latest" instead of a specific number in that field, or we must update the value now and again.

umairidris:
Note https://github.com/terraform-google-modules/terraform-google-kubernetes-engine uses "latest" as default for "kubernetes_version".

PR #241, opened 2019-07-17
Enable GKE DM template to support 'latest' initial cluster version.

gruihuang:
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/177
Do not get "clientCertificate" and "clientKey" when "latest" is used in the initialClusterVersion.

### DIFF
--- a/dm/templates/gke/gke.py
+++ b/dm/templates/gke/gke.py
@@ -14,7 +14,6 @@
 """ This template creates a Google Kubernetes Engine cluster. """
 
 from packaging import version
-import re
 
 def generate_config(context):
     """ Entry point for the deployment resources. """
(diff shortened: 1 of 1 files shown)


dm/templates/gke/gke.py
```diff
@@ -116,15 +117,17 @@ def generate_config(context):
...
-            propc.get('masterAuth', {}).get('clientCertificateConfig', False)
-        )
-    ):
-        output_props.append('clientCertificate')
-        output_props.append('clientKey')
+    initial_cluster_version = propc.get('initialClusterVersion')
+    if initial_cluster_version.lower() != 'latest':
+        if (
+            # https://github.com/GoogleCloudPlatform/deploymentmanager-samples/issues/463
+            propc.get('enableDefaultAuthOutput', False) and (
```
morgante: This doesn't look quite right.

According to the existing logic, if I have "latest" (ie >= 0.12) *and* propc.get('masterAuth', {}).get('clientCertificateConfig', False), then we should still append these output props.

Could you update this to make the logic on this clearer. Instead of putting everything in a large block, I'd suggest creating a greater_than_012 conditional (which is also true on latest).

gruihuang: You are right, done.

PR #245, opened 2019-07-18
WIP: Update lint image with task helper function

glarizza:
This commit updates the CFT lint image with the helper functions
previously saved as `test/make.sh` in every module. These helper
functions are being centralized within the image to reduce code
duplication across the modules. The file has been named to
`task_helper_functions.sh` to indicate that these helper functions are
available for module tasks regardless of whether those tasks
are called from `make` or not.

The commit also upgrades the version of Terraform to 0.12.4 and bumps
the minor version of the lint image to accommodate the change.


glarizza:
This PR is a bit premature, closing for now

PR #249, opened 2019-07-22
turn billing account info on project creation

tyron:
providing a billing account info is optional for project creation. if the developer is not a billing account admin, they should still be able to create projects to try out CFT.


ocsig:
Have you tested all the other functionalities of the template without the billing account attached? For example creation of usage export bucket or enabling APIs?

ocsig:
Hi Tryon,

We really appriciate your contribute, please don't stop! However we do not wish to merge your changes to our codebase because of the following reasons:
- Creating a project without a billing account is a really edge case since that project will not be usable at that stage.
- Our current Project template supports multiple usefull features such as enabling services, shared VPCs. These would fail without a billing account.

Reflecting on your reasoning:

> If the developer is not a billing account admin, they should still be able to create projects to try out CFT

If the developers can't get access to a dummy bulling account, they can't fully test project creation. However they are able to test the other templates of CFT in an existing project with an attached billing account.

PR #251, opened 2019-07-23
Add support for Terraform 0.12 to Event Function pipeline

### LINKED ISSUE #20: Project template fails with removeDefaultVPC or removeDefaultSA is True if compute api is not enabled
We should skip removing VPC if compute api is not in the list

bohdanyurov-gl:
https://github.com/terraform-google-modules/terraform-google-event-function/issues/20

Updated CI pipeline for Event function


aaron-lane:
@nick4fake and @bohdanyurov-gl are the same person.

PR #257, opened 2019-07-29
Isolate a single template from route DM template.

gruihuang:
Currently, the route DM template creates a group of routes instead of a single route. Sometimes we want users to create only one resource once.
This PR does not change the function of "route" template. It moves the logic that creates a route into a new template "gce_route.py". And the "route.py" uses the new template to deploy routes.


dm/templates/route/single_route.py
```diff
@@ -0,0 +1,118 @@
...
+    ]
+
+    for prop in optional_properties:
+        if prop in properties:
+            res_properties[prop] = properties[prop]
+
+    name = properties['name']
+    resources = [
+        {
+            'name': name,
```
nick4fake: Please use context.env['name'] and add name to properties

nick4fake: You have two names here:
- DM resource name
- real resource name

Previously we were using real resource names as DM resource names, that is incorrect. DM resource name should always be context['name'], while real resource name can be overridden by user

nick4fake: I don't see properties->name being set

PR #281, opened 2019-08-15
accept local directory path as input

katze120:
accept local directory path as input


katze120:
resubmitting a new one shortly.

PR #304, opened 2019-09-04
Give CI org admin role on test org

morgante:


infra/terraform/test-org/org/iam.tf
```diff
@@ -57,3 +67,9 @@ module "ci_folders_folder_bindings" {
     ]
   }
 }
+
+resource "google_billing_account_iam_member" "ci-billing-user" {
+  billing_account_id = local.billing_account
+  role               = "roles/billing.user"
```
paulpalamarchuk: isn't it "roles/billing.admin" ?

PR #306, opened 2019-09-04
Add script for deleting all projects within a folder

morgante:


infra/utils/delete-projects.py
```diff
@@ -0,0 +1,81 @@
...
+        d_request = service.liens().delete(name=lien.get('name'))
+        d_request.execute()
+        liens_deleted += 1
+
+    return liens_deleted
+
+def delete_project(project):
+    try:
+        project.delete()
+    except google.api_core.exceptions.BadRequest as e:
```
aaron-lane: This exception should be printed for clarity, either conditionally if no liens are deleted or unconditionally with a note that lien deletion will be attempted.

morgante: @aaron-lane Fixed, PTAL.

PR #307, opened 2019-09-04
Launchpad folder generation

mikeweng:
Generation of Terraform based GCP Folders based on CloudFoundation/Organization/Folder CRD input.


mikeweng:
Thank you @dparrish for all the readability suggestions!

I'll be updating this code for Terraform 12.

cli/launchpad/functionality.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+)
+
+// directoryOwner interface allows implementers to specify directory creation.
```
morgante: This `directory` construct is somewhat confusing to me.

cli/launchpad/model.go
```diff
@@ -0,0 +1,248 @@
...
+	rawYAML string
+}
+
+// commonConfigYAML alias configYAML for all CRDs to implement common properties.
+type commonConfigYAML configYAML
+
+// UnmarshalYAML evaluates common CRD attributes and dynamically parse the CRDs based on Kind field value.
+func (c *configYAML) UnmarshalYAML(unmarshal func(interface{}) error) (err error) {
+	type raw configYAML
+	if err := unmarshal((*raw)(c)); err != nil {
```
dparrish: What is the purpose of the `type raw` here? `unmarshal` already takes an interface{} so you can provide `c` to it directly

```
if err := unmarshal(c); err != nil {
```

cli/launchpad/model.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+// configYAML represents a fully qualified CRD that can be of any supported Kind.
+//
+// All CRDs are expected to have the following properties.
+type configYAML struct {
+	APIVersion string      `yaml:"apiVersion"`
+	Kind       crdKind     `yaml:"kind"`
+	Spec       interface{} `yaml:"spec"` // Placeholder, yielding for further evaluation
+	//Metadata   Metadata               `yaml:"metadata"`
```
dparrish: Remove if it's unused.

cli/launchpad/model.go
```diff
@@ -0,0 +1,248 @@
...
+	Undefined map[string]interface{} `yaml:",inline"` // Catch-all for untended behavior
+
+	// Original YAML for backup purpose
+	rawYAML string
+}
+
+// commonConfigYAML alias configYAML for all CRDs to implement common properties.
+type commonConfigYAML configYAML
```
dparrish: Is this necessary? You seem to just be mixing configYAML and comonConfigYAML in odd places.

cli/launchpad/model.go
```diff
@@ -0,0 +1,248 @@
...
+
+// UnmarshalYAML evaluates common CRD attributes and dynamically parse the CRDs based on Kind field value.
+func (c *configYAML) UnmarshalYAML(unmarshal func(interface{}) error) (err error) {
```
dparrish: You shouldn't name this return value

```
func (c *configYAML) UnmarshalYAML(unmarshal func(interface{}) error) error {```

File: cli/launchpad/model.go [hunk 6]
```diff
@@ -0,0 +1,248 @@
...
+	type raw configYAML
+	if err := unmarshal((*raw)(c)); err != nil {
+		return err
+	}
+	switch c.Kind {
+	case KindCloudFoundation:
+		cf := cloudFoundationYAML{}
+		cf.APIVersion = c.APIVersion
+		cf.Kind = c.Kind
+		err = unmarshal(&cf)
```
dparrish: I'm very confused by this. I think you're mixing up the marshal / unmarshal terms.
https://en.wikipedia.org/wiki/Marshalling_(computer_science)

If this is really doing an unmarshal, it should be taking the serialized form of something and turning it into the inmemory representation, which in this case is a `cloudFoundationYAML` struct.

To be clear:
"Marshal" - Turn an in-memory struct into YAML
"Unmarshal" - Turn YAML into an in-memory struct.

I guess it's possible that the `unmarshal` func passed to UnmarshalYAML has some data reference in the closure but I can't find any uses of it to see.

This can also return the err directly.

cli/launchpad/model.go
```diff
@@ -0,0 +1,248 @@
...
+		f.APIVersion = c.APIVersion
+		f.Kind = c.Kind
+		err = unmarshal(&f)
+	case KindOrganization:
+		o := orgYAML{}
+		o.APIVersion = c.APIVersion
+		o.Kind = c.Kind
+		err = unmarshal(&o)
+	default:
+		return errors.New(fmt.Sprintf("Kind %s not implemented", c.Kind))
```
dparrish:
```
return fmt.Errorf("Kind %s not implemented", c.Kind)
```

cli/launchpad/model.go
```diff
@@ -0,0 +1,248 @@
...
+	Spec folderSpecYAML `yaml:"spec"`
+}
+
+// folderSpecYAML defines Folder Kind's spec.
+type folderSpecYAML struct { // Inner mappings
+	Id          string                 `yaml:"id"`
+	DisplayName string                 `yaml:"displayName"`
+	ParentRef   parentRefYAML          `yaml:"parentRef"`
+	Folders     []*folderSpecYAML      `yaml:"folders"`
+	Undefined   map[string]interface{} `yaml:",inline"` // Catch-all for untended behavior1
```
dparrish: "untended" is not a word.

cli/launchpad/model.go
```diff
@@ -0,0 +1,248 @@
...
+// UnmarshalYAML will have side effect to push Folder onto its stack while nested
+// evaluation is ongoing. In addition, storing validated folder onto gState for tracking.
+//
+// GCP Folder names required to be in between 3 to 30 characters
+func (f *folderSpecYAML) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	top := gState.peek()
+	gState.push(KindFolder, f)
+	defer gState.popSilent()
+
+	type raw folderSpecYAML
```
dparrish: I keep seeing this pattern in this code. I'm very sure it's wrong. See above for why.

cli/launchpad/model_test.go
```diff
@@ -0,0 +1,155 @@
...
+		folderSpecYAML{
+			Id:          "group1_2",
+			DisplayName: "Group 1-2",
+			ParentRef: parentRefYAML{
+				"Folder",
+				"group1",
+			},
+		},
+	}
+	//folderG21 := &folderYAML{
```
dparrish: Remove if not used.

cli/launchpad/model_test.go
```diff
@@ -0,0 +1,155 @@
...
+	}}
+
+	for _, tc := range testCases {
+		tt.Run(tc.name, func(t *testing.T) {
+			evalYAMLs := &gState.evaluated.folders.YAMLs // Evaluated YAMLs
+
+			defer func() { *evalYAMLs = make(map[string]*folderSpecYAML) }()
+			err := loadAllYAMLs(tc.filePatterns)
+
+			if err != nil && tc.loadErr != nil && err.Error() != tc.loadErr.Error() {
```
dparrish: Testify has `require.EqualError()` for this purpose.

cli/launchpad/model_test.go
```diff
@@ -0,0 +1,155 @@
...
+	}, {
+		"folder_under_org_CRD",
+		nil,
+		testdataFiles("folder/org_1.yaml"),
+		map[string]*folderYAML{"group1_2": &folderOrgG12},
+	}}
+
+	for _, tc := range testCases {
+		tt.Run(tc.name, func(t *testing.T) {
+			evalYAMLs := &gState.evaluated.folders.YAMLs // Evaluated YAMLs
```
dparrish: I'm concerned by this.

You probably shouldn't ever be creating a pointer to a slice. If you're doing that, and that requires you to dereference it to check the length, you're very likely doing it wrong.

cli/launchpad/root.go
```diff
@@ -0,0 +1,85 @@
...
+	gState.references = make(map[string]*parentRefYAML)
+	gState.evaluated.folders.YAMLs = make(map[string]*folderSpecYAML)
+}
+
+// NewGenerate takes file patterns as input YAMLs and output Infrastructure as
+// Code ready scripts based on specified output flavor.
+//
+// NewGenerate can be triggered by
+//   $ cft launchpad generate
+func NewGenerate(rawFilepath []string, outFlavor string, outputDir string) {
```
dparrish: I don't personally see the need for defining `type outputFlavor` above, string is probably fine.

But if you define it then you should use it. You're taking `outFlavor` as string here, when it probably should be `outputFlavor`.

cli/launchpad/root.go
```diff
@@ -0,0 +1,85 @@
...
+	"log"
+)
+
+//go:generate go run static/includestatic.go
+
+// crdKind is the CustomResourceDefinition (CRD) which is indicated YAML's Kind value.
+type crdKind string
+
+// outputFlavor defines launchpad's generated output language.
+type outputFlavor string
```
dparrish: This looks like it's being used as a enum. You should make it an enum then.

```go
type outputFlavor int
const (
   outDM outputFlavor = iota
   outTF
)
```

Same for crdKind.

cli/launchpad/root.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+	if fps == nil || len(fps) == 0 {
+		return errors.New("no valid YAML files given")
+	}
+	for _, conf := range fps { // Load all files into runtime
+		// TODO multiple yaml documents in one file
+		if content, err := loadFile(conf); err != nil {
+			return err
+		} else {
```
dparrish: `else` here is not required. `return err` above will stop this block running if there's an error. Go style encourages de-intending code as much as possible

```
content, err := loadFile(conf)
if err != nil {
  return err
}
// Continue along with your business.
```

cli/launchpad/runtime.go
```diff
@@ -0,0 +1,154 @@
...
+	references      map[string]*parentRefYAML
+	evaluated       struct {
+		orgId   string
+		folders folders
+	}
+}
+
+// ==== Evaluation Stack ====
+
+// stackable interface determines what can be pushed onto the stack.
```
dparrish: This is *not* defining an interface. This is just creating a new type that is the same as an existing type.

A stackable interface might look something like:

```go
type stackable interface {
  Unmarshaler([]byte) error
}
```

That means that every object that implements the `Unmarshaler` method described will implicitly implement the `stackable` interface.

What you have here is *not* an interface in any way and is just over-complicating the code with the illusion of making it more flexible.

cli/launchpad/intputs.go
```diff
@@ -0,0 +1,61 @@
+// Package launchpad file inputs.go contains all input processing logic.
```
morgante: File name has typo.

cli/launchpad/model.go
```diff
@@ -0,0 +1,310 @@
...
+
+// UnmarshalYAML evaluates common CRD attributes and dynamically parse the CRDs based on CRD Kind.
+func (g *genericYAML) UnmarshalYAML(unmarshal func(interface{}) error) (err error) {
+	type raw genericYAML
+	if err := unmarshal((*raw)(g)); err != nil {
+		return err
+	}
+	switch g.Kind() {
+	case CloudFoundation:
+		return unmarshal(&struct {
```
morgante: As discussed, I think we will want to return the full CRD (not just the spec).

cli/launchpad/model.go
```diff
@@ -0,0 +1,310 @@
...
+	type raw referenceYAML
+	if err := unmarshal((*raw)(r)); err != nil {
+		return err
+	}
+
+	gState.referenceMap.putExplicit(src, r) // retain explicit reference to check if reference exist
+	return nil
+}
+
+// ==== CloudFoundation ====
```
morgante: I think we might want to split each type into a separate file?

cli/launchpad/root.go
```diff
@@ -0,0 +1,59 @@
...
+
+// init initialize tracking for evaluated objects
+func init() { gState.init() }
+
+// NewGenerate takes file patterns as input YAMLs and output Infrastructure as
+// Code ready scripts based on specified output flavor.
+//
+// NewGenerate can be triggered by
+//   $ cft launchpad generate
+func NewGenerate(rawPaths []string, outputFlavorString string, outputDir string) {
```
morgante: I think we should accept output flavor as the type directly (not a string).

cli/launchpad/generic.go
```diff
@@ -0,0 +1,106 @@
...
+
+const indentSize = 2
+
+var (
+	errValidationFailed = errors.New("validation failed")
+	tfNameRegex         = regexp.MustCompile("^[a-zA-Z][a-zA-Z\\d\\-\\_]*$")
+)
+
+//
+type resourcer interface {
```
morgante: We shouldn't call this `resourcer` as `resource` is not a verb. Perhaps `ResourceHandler`?

cli/launchpad/generic.go
```diff
@@ -0,0 +1,106 @@
...
+	// validate ensures the parsed YAML has validate parameters.
+	validate() error
+	// kind returns the validated crdKind of the resource.
+	kind() crdKind
+	// enroll adds the resource into final eval mapping.
+	enroll(*eval) error
+}
+
+type resourcespecer interface {
```
morgante: Having this separate interface feels wrong to me. I'd see if we can consolidate everything into the top-level interface.

cli/launchpad/runtime.go
```diff
@@ -0,0 +1,137 @@
...
+// for evaluation hierarchy and eval object tracking.
+package launchpad
+
+import (
+	"fmt"
+	"log"
+	"strings"
+)
+
+type eval struct {
```
morgante: Let's have a descriptive name for this. Is it a `config`? Or an `orgMap`?

cli/launchpad/runtime.go
```diff
@@ -0,0 +1,137 @@
...
+	rmap referenceMap
+	org  orgSpecYAML
+}
+
+func evaluate(rs []resourcer) *eval {
```
morgante: Let's use more descriptive function names. Maybe `buildHierarchy`?

cli/launchpad/runtime.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// referenceMap maps referenced target to a list of source references.
+type referenceMap struct {
+	mappings map[string][]reference
+	ids      map[string]resourcespecer
+}
+
+// organize loop over all addressable target and ensure parent child relationships.
+func (m *referenceMap) organize() error {
```
morgante: Is this more accurately called `resolveReferences`? `organize` is quite generic.

cli/launchpad/runtime.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+type eval struct {
+	rmap referenceMap
+	org  orgSpecYAML
+}
+
+func evaluate(rs []resourcer) *eval {
+	e := eval{}
+	e.rmap.ids = make(map[string]resourcespecer)
```
morgante: Why can't these be the full resources (instead of just the specs)?

cli/launchpad/org.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/pkg/errors"
+	"strings"
+)
+
+type orgYAML struct {
+	headerYAML `yaml:",inline"`
+	Spec       orgSpecYAML `yaml:"spec"`
+}
+
+func (o *orgYAML) enroll(e *eval) error { return o.Spec.enroll(e) }
```
morgante: These wrapping funcs definitely smell like unnecessary code to me. Let's just operate entirely on the top level (or entirely on the specs).

cli/launchpad/folder.go
```diff
@@ -0,0 +1,106 @@
...
+	e.register(f, &f.ParentRef)
+	for _, sf := range f.Folders { // Recursively enroll sub-folders
+		if err := sf.enroll(e); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (f *folderSpecYAML) resolveRefs(refs []reference) error {
```
morgante: Would it be more accurate to call this `attachChildren`? Or even `attachReferences` if we want to be generic?

cli/launchpad/folder.go
```diff
@@ -0,0 +1,145 @@
...
+	folderNameMax = 30
+)
+
+// folderSpecYAML defines GCP Folder's spec.
+type folderSpecYAML struct { // Inner mappings
+	Id             string                 `yaml:"id"`
+	DisplayName    string                 `yaml:"displayName"`
+	ParentRef      referenceYAML          `yaml:"parentRef"`
+	SubFolderSpecs []*folderSpecYAML      `yaml:"folders"`
+	Undefined      map[string]interface{} `yaml:",inline"` // Catch-all for untended behavior
```
morgante:
```suggestion
	Undefined      map[string]interface{} `yaml:",inline"` // Catch-all for unintended behavior
```

cli/launchpad/folder.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	switch f.Spec.ParentRef.TargetTypeStr { // Validate Supported Parents
+	case Organization.String(), Folder.String():
+	default:
+		log.Printf("fatal: unsupported parent '%s' type for Folder\n", f.Spec.ParentRef.TargetTypeStr)
+		return errInvalidParent
+	}
+
+	// TODO validate misc
+	if ind := tfNameRegex.FindStringIndex(f.Spec.Id); ind == nil {
```
morgante: Shouldn't this just be `MatchString`? We don't need the index.

cli/launchpad/folder.go
```diff
@@ -0,0 +1,145 @@
...
+		log.Printf("GCP Folder [%s] ID does not conform to Terraform standard", f.Spec.DisplayName)
+		return errValidationFailed
+	}
+
+	if len(f.Spec.DisplayName) < folderNameMin || len(f.Spec.DisplayName) > folderNameMax {
```
morgante: Technically length limits could be included in regex (just FYI).

cli/launchpad/folder.go
```diff
@@ -0,0 +1,145 @@
...
+			f.subFolders.add(r)
+		default:
+			log.Printf("fatal: invalid %s parent for %s\n", f.refId(), r.refId())
+			return errInvalidInput
+		}
+	}
+	return nil
+}
+
+// dump generates debug string slices representation.
```
morgante: What if instead of returning an array of strings and incrementally adding to it, how about `dump` accepts an `io.Writer` and writes to it directly?

(Note that if we *do* want it as a string later you could always pass in, `strings.Builder`).

cli/launchpad/folder.go
mikeweng: This is checking a different field, regex check for id, this check for DisplayName.

cli/launchpad/folder.go
mikeweng: Awesome idea, added in

morgante: Looks like the comment still needs to be updated.

morgante:
@mikeweng Thanks for your patience, merged!

PR #312, opened 2019-09-06
[DM] Cloudbuild triggers for CI

ocsig:


ocsig:
@sjvanrossum Note, the project test has known cleanup issues, the rest is passing.

PR #314, opened 2019-09-06
Scorecard - allow input of CAI files from local filesystem, and take all files in the given bucket or local directory

katze120:
Scorecard feature updates

- Allow input of CAI files from local filesystem

- Take all files in the given bucket or local directory
Just like Forseti supports 'composite_root_resources', this feature will enable aggregated analysis for iam and resource CAI export files from multiple folders and projects at the same file.


cli/scorecard/violations.go
```diff
@@ -67,20 +95,40 @@ func addDataFromBucket(config *ScoringConfig, bucket *storage.BucketHandle, obje
 func getViolations(inventory *InventoryConfig, config *ScoringConfig) (*validator.AuditResponse, error) {
 	v := config.validator
 
-	ctx := context.Background()
-	client, err := storage.NewClient(ctx)
-	if err != nil {
-		return nil, err
-	}
+	if inventory.gcsBucket != "" {
```
morgante: Instead of using an `if` block here, I'd suggest we refactor this slightly (especially in light of duplication in `addDataFromBucket` and `addDataFromFile`). I'd suggest we have:

1. getViolations calls either `addDataFromBucket` or `addDataFromDirectory`)
2. `addDataFromBucket` and `addDataFromDirectory` contain most of the logic currently implemented here and both call `addDataFromReader`.

cli/scorecard/inventory.go
```diff
@@ -52,10 +53,11 @@ func TargetProject(projectID string) Option {
 }
 
 // NewInventory creates a new CAI inventory manager
-func NewInventory(projectID string, bucketName string, options ...Option) (*InventoryConfig, error) {
+func NewInventory(projectID string, bucketName string, caiDirName string, options ...Option) (*InventoryConfig, error) {
```
morgante: Instead of adding new, potentially empty, params we should probably refactor `bucketName` and `caiDirName` to be Options.

cli/scorecard/violations.go
```diff
@@ -48,6 +50,32 @@ func addDataFromBucket(config *ScoringConfig, bucket *storage.BucketHandle, obje
...
+	if err != nil {
+		return errors.Wrap(err, "Fetching inventory")
+	}
+
+	return nil
+}
+
+func addDataFromFile(config *ScoringConfig, objectName string) error {
+	reader, err := os.Open(objectName)
+	if err != nil {
```
morgante: Looks like everything beyond this line is identical in `addDataFromBucket`. Could we consolidate on `addDataFromReader`?

cli/scorecard/violations.go
```diff
@@ -67,20 +95,40 @@ func addDataFromBucket(config *ScoringConfig, bucket *storage.BucketHandle, obje
...
+		client, err := storage.NewClient(ctx)
+		if err != nil {
+			return nil, err
+		}
 
-	bucket := client.Bucket(inventory.gcsBucket)
-	for _, objectName := range destinationObjectNames {
-		err = addDataFromBucket(config, bucket, objectName)
+		bucket := client.Bucket(inventory.gcsBucket)
+		it := bucket.Objects(ctx, nil)
```
morgante: This is something of an interface change which I'd like to consider in a separate PR. In particular, I'm not entirely sure want to switch to scanning all objects in the bucket instead of well-known ones. Could you back this out?

cli/scorecard/inventory.go
katze120: addressed in commit https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/314/commits/dfb1e39be1eaff427d9e21f81df5bb8ff87f6352

cli/scorecard/violations.go
katze120: addressed in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/314/commits/04fb6f7898a992131c4801e1c2e5d5996eeff114

cli/scorecard/violations.go
katze120: addressed in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/314/commits/04fb6f7898a992131c4801e1c2e5d5996eeff114

cli/scorecard/violations.go
katze120: Backed out in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/314/commits/7801912bc8ac86403880f8b5dcaa73c6e1c9e05b

I'm hoping the tool can take multiple resource and iam files at a time because Forseti has `composite_root_resources` field in server config which allows users to scan multiple org/folders/projects at the same time, the CAI exports generated by Forseti have unique file names such as `projects-<project_id>-iam_policy-<timestamp>.dump` and `projects-<project_id>-resource-<timestamp>.dump`
One use case I'm hoping this tool to support would be running the tool against the latest CAI export files in Forseti CAI bucket.
On local file systems it's relatively easy to create sub folders for each node (org, folder, or project) resource and organize CAI export files with expected file names into sub folders, then run a for loop. On GCS it's more difficult as it involves creating additional buckets or some additional code change.

cli/scorecard/cmd.go
```diff
@@ -20,11 +21,14 @@ func init() {
 
 	Cmd.Flags().StringVar(&flags.targetProjectID, "project", "", "Project to analyze (conflicts with --organization)")
 
-	Cmd.Flags().StringVar(&flags.bucketName, "bucket", "", "GCS bucket name for storing inventory")
-	Cmd.MarkFlagRequired("bucket")
+	Cmd.Flags().StringVar(&flags.inputPath, "input-path", "", "GCS bucket name (by default) OR local directory path (with --input-local option), for storing inventory")
```
morgante: I think the way you had it before is better actually, having the same argument refer to two different options seems undesirable and harder to validate. Could we stick to two options: bucket or file-path?

cli/scorecard/violations.go
morgante: Agreed, happy to look at a PR which separately does that. We'll just need to consider whether it should scan *all* the files or only a subset (to capture the latest dump).

katze120: Ideally only a subset (to capture the latest dump), but the file names do not make it straight forward to extract those. Agreed to revisit this in a separate PR.

cli/scorecard/cmd.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/314/commits/68d7ddb5bf241cef64e2b2a0b42fe9f1a7472b41
(I call the 2nd param `dir-path` instead of `file-path` to indicate this parameter takes a directory)

PR #315, opened 2019-09-06
Inventory reporting feature under new 'report' subcommand

katze120:


cli/report/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+
+	Cmd.Flags().StringVar(&flags.queryPath, "query-path", "", "Path to directory containing inventory queries")
+	Cmd.MarkFlagRequired("query-path")
+
+	Cmd.Flags().StringVar(&flags.reportPath, "report-path", "", "Path to directory to contain report outputs")
+
+	Cmd.Flags().StringVar(&flags.bucketName, "bucket", "", "GCS bucket name for storing inventory (conflicts with --local-cai-path)")
+	Cmd.Flags().StringVar(&flags.caiDirName, "local-cai-path", "", "Local directory path for storing inventory (conflicts with --bucket)")
+
+	Cmd.Flags().StringVar(&flags.reportFormat, "format", "", "Format of inventory report outputs, can be json or csv, default is csv")
```
morgante: We might want to use Viper to manage these flags, as it includes default-setting functionality: https://github.com/spf13/viper#establishing-defaults

cli/report/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	
+	Cmd.Flags().BoolVar(&flags.listReports, "list-available-reports", false, "List available inventory report queries")
```
morgante: This should be implemented as a different command, (`cft report list` ?) rather than a flag on the root command.

cli/report/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	listReports  bool
+}
+
+func init() {
+	viper.AutomaticEnv()
+
+	Cmd.Flags().StringVar(&flags.queryPath, "query-path", "", "Path to directory containing inventory queries")
+	Cmd.MarkFlagRequired("query-path")
+
+	Cmd.Flags().StringVar(&flags.reportPath, "report-path", "", "Path to directory to contain report outputs")
```
morgante: This should be required, right? Also I think we should make it more explicit that it's for outputs. Maybe `output-path` or `report-output-path`.

cli/report/cmd.go
```diff
@@ -0,0 +1,88 @@
...
+	Example:
+	  cft report --query-path ./path/to/cloud-foundation-toolkit/cli/samplereports \
+		--local-cai-path ./path/to/cai-export-directory \
+		--report-path ./path/to/report-output-directory \
+	`,
+	
+	Args:  cobra.NoArgs,
+	PreRunE: func(c *cobra.Command, args []string) error {
+		if !flags.listReports {
+			if flags.reportPath == "" {
```
morgante: We should handle this via Cobra: https://github.com/spf13/cobra#required-flags

cli/report/convert.go
```diff
@@ -0,0 +1,76 @@
...
+	"encoding/json"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+// ReadFilesAndConcat reads json files in a directory that are one object per row,
+// and concats all objects into one single array
+func ReadFilesAndConcat(dir string) (results []interface{}, err error) {
```
morgante: Note this approach isn't very efficient (needing to load all objects into an array *before* report generation). It's not a blocker for this PR, but in future iterations we're going to have to explore streaming input/output.

Also not sure why we need to export this function? Seems internal.

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+	"strings"
+	"time"
+
+	"github.com/open-policy-agent/opa/loader"
+	"github.com/open-policy-agent/opa/rego"
+)
+
+// GenerateReports takes CAI output and generate inventory report content based on report queries
+func GenerateReports(caiPath string, reportQueryPath string, reportOutputPath string, reportFormat string) error {
+	fileSuffix := time.Now().Format("20060102150405")
```
morgante: Could we make this suffix more readable?

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/open-policy-agent/opa/loader"
+	"github.com/open-policy-agent/opa/rego"
+)
+
+// GenerateReports takes CAI output and generate inventory report content based on report queries
+func GenerateReports(caiPath string, reportQueryPath string, reportOutputPath string, reportFormat string) error {
```
morgante: nit. It'd be good to have clearer documentation on the expected inputs/outputs *or* don't export this function at all.

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+	fileSuffix := time.Now().Format("20060102150405")
+	rawAssetFileName, err := convertAndGenerateTempAssetFile(caiPath, reportOutputPath, fileSuffix)
```
morgante: This approach seems very inefficient. I'm not quite sure the necessity of generating a temp asset file?

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+	rs, err := r.Eval(context.Background())
+	if err != nil {
+		return nil, err
+	}
+	results = rs[0].Expressions[0].Value
+
+	return results, nil
+}
+
+func printReports(results interface{}, reportOutputPath string, format string, fileSuffix string) error {
```
morgante: We might want to consider typing `results` more strongly? Or cast it into a map earlier.

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+	resultsMap := results.(map[string]interface{})
+	for group, reports := range resultsMap {
+		reportsMap := reports.(map[string]interface{})
```
morgante: These aren't all reports though? Maybe consider a more descriptive name. `groupItems` or something generic?

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+		for reportName, content := range reportsMap {
+			if strings.HasSuffix(reportName, "_report") {
+				reportFileName := group + "." + reportName + "_" + fileSuffix
```
morgante: We might use fmt.sprintf instead?

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+				fmt.Println("Printing " + group + "." + reportName)
```
morgante: This should use fmt.Printf instead.

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+							recordMap := record.(map[string]interface{})
+							var record []string
+							for _, key := range keys {
+								record = append(record, recordMap[key].(string))
+							}
+							w.Write(record)
+						}
+						w.Flush()
+					} else {
+						f, err := os.Create(filepath.Join(reportOutputPath, reportFileName))
```
morgante: I'm as little confused on the logic here. Can we explain more? It seems like this line is recreating the same file from line 111?

cli/report/report.go
```diff
@@ -0,0 +1,184 @@
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func ListAvailableReports(reportQueryPath string) error {
+	// Load resources from json and rego files
+	resources, err := loader.All([]string{reportQueryPath})
```
morgante: This logic looks duplicative, I'd make a small suggestion to refactor so this and `generateReportData` share common underlying methods. In particular, logic like finding reports and knowing their suffix shouldn't be duplicated.

Non-blocking.

cli/samplereports/README.md
```diff
@@ -0,0 +1,4 @@
+This is the folder for report queries.
```
morgante: I think we can put this at the root (ie `github.com/GoogleCloudPlatform/cloud-foundation-toolkit/reports/sample`).

cli/report/cmd.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/8d47e44125e9ad0d4c53bf2b7f88642ccdd9b286

cli/report/cmd.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/7b81a27b8c8a7f699312924d16a1ef23d1c985fb

cli/report/cmd.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/63c45f8531ac18c0e392f9b2fd08fad7ec5e0fcb

cli/report/cmd.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/63c45f8531ac18c0e392f9b2fd08fad7ec5e0fcb

cli/report/convert.go
katze120: This function is called by `convertAndGenerateTempAssetFile` in report.go hence exported.

Here are reasons to load all objects into array first: 

- The resultant file can be directly feed into [opa loader.All()](https://github.com/open-policy-agent/opa/blob/master/loader/loader.go#L82-L84), to avoid additional coding to implement similar function. 
- The resultant file is a human readable json file which is also being written to `output-path` as `raw_assets_<timestamp>.json`, I find useful for 1) ad-hoc inventory inspection, and 2) feeding into `opa` command directly for rego development and troubleshooting (Our policy-library test fixtures also have input asset files in one json array, while CAI exports are in the format of one json object per line)

Agreed that this approach is not ideal for performance. If we have a streaming solution, I would have the `report` command to use streaming approach and move the generation of one json array file to a separate feature.

cli/report/report.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/3cb59d828e7fcd0b82192f1381ab6411a2fcba1a

cli/report/report.go
katze120: explained in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315#discussion_r329341299

cli/report/report.go
katze120: Function needs to be exported because it's used by cmd.go

Updated function description and parameter names in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/2c5e2414d3ca6cf6f707b03a5a007b491ad49daf

cli/report/report.go
katze120: The reports do not have fixed number of fields or fixed field names so I'm not sure how to do this. Would you please elaborate ?

cli/report/report.go
katze120: As of now these things are all inventory reports
A "group" can be "iam", "vm", "network", "data" etc,
and a "report" can be "iam.bindings_report", "vm.service_account_report", "network.subnet_flow_logs_report", "data.cloud_sql_public_authorized_networks_report", etc

cli/report/report.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/0c75b7e2cbd6c6c9809f57aa9c015b70d80f711d

cli/report/report.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/0c75b7e2cbd6c6c9809f57aa9c015b70d80f711d

cli/report/report.go
katze120: removed these lines in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/136717b50a018bb8612f5f81f51d7c4c173c5c2a
good catch! (both this one and the rest) I can't recall why I did this.

cli/samplereports/README.md
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/ebdf1fbc6190f26175d669b6b90d106f8c5bcd5f

cli/report/report.go
katze120: done in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/315/commits/df61d7d7fd5592a35b28cfc9d0f10afb533a3aaf

cli/report/convert.go
morgante: Alright, that sounds reasonable to me.

cli/report/report.go
morgante: Well, specifically results must be `map[string]interface{}` so it'd make sense to simply accept that as the input IMO.

cli/report/report.go
morgante: Got it.

cli/report/report.go
morgante: I actually meant to use fmt.sprintf for forming the file name instead of concatenation, but it's just a minor suggestion.

PR #317, opened 2019-09-09
Fixes #274: Invalid example - Cloud build trigger

### LINKED ISSUE #274: Invalid example - Cloud build trigger
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/10ab68c1680f61a3a1bfd929dbba760f216f952e/dm/templates/cloudbuild/examples/cloudbuild_trigger.yaml#L30

Property 'build' got renamed/moved in the schema, needs a fix.

nick4fake:
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/274

Fixed example and schema


dm/templates/cloudbuild/cloudbuild.py.schema
```diff
@@ -412,12 +412,12 @@ definitions:
           The directory to run the build in (relative to the source root).
           If a step's dir is specified as an absolute path, the value is
           ignored for that step's execution.
-      revision:
```
ocsig: This needs to stay 'revision'. If you call the property same as the replacements inside the oneOfs it fails

nick4fake: I can not call it, as this "revision" field is not supported in gcp type.

PR #322, opened 2019-09-09
Adding ProjectNumber to project template output and schema

ocsig:


ocsig:
Note: Project test deletion fails because of known issue.

Issue #299, opened 2019-08-27
Template for unmanaged instance group


ocsig:
#288 Creation of a template, simmillare to the MIG template, to support Unamanaged instance groups.

PR #328, opened 2019-09-12
Create template for the UMIG

paulpalamarchuk:
* Added UMIG template (feature #299)


paulpalamarchuk:
@googlebot I signed it!

paulpalamarchuk:
Tests:
image

dm/templates/unmanaged_instance_group/unmanaged_instance_group.py
```diff
@@ -0,0 +1,128 @@
...
+                project_id,
+                zone,
+                instance_name.get('instance')
+            )
+            instance = { 'instance': instance_url }
+            formatted_instances.append(instance)
+
+        add_instances = {
+            'name': '{}-addinstances'.format(name),
+            'action': (instance_provider_type),
```
ocsig: This will try to add every instance at every update. This needs to have a runtime policy for create only

paulpalamarchuk: @ocsig 
I can add runtime policy and change input structure like that:
```
instances:
  add: 
  - instance-1
  - instance-2
  remove:
  - instance-3
```
Is that acceptable?

ocsig:
Please test the following scenarios:
- Creation with instance 1, 2 instances.
- Update with instance 1,2,3 >> adding instance 3 only
- Update with instance 3,1,2 >> No-operatio
- Update with instance 2,3 >> removing instance 1

PR #354, opened 2019-09-26
Add bats-support, bats-mock, bats-assert libraries

averbuks:
Additional bats libraries are required for testing terraform-google-modules/terraform-google-startup-scripts#28


infra/concourse/build/developer-tools/Dockerfile
```diff
@@ -73,6 +85,15 @@ RUN /build/install_terraform_docs.sh ${TERRAFORM_DOCS_VERSION}
 ADD ./build/install_bats.sh /build/
 RUN /build/install_bats.sh ${BATS_VERSION}
 
+ADD ./build/install_bats_support.sh /build/
```
morgante: Nit: Could we just combine these into a single `install_bats.sh` script?

averbuks: Good point, addressed.

PR #356, opened 2019-09-27
Only add test-base-path to main kitchen commands

aaron-lane:
This branch patches `kitchen_do` to only add `--test-base-path` to the relevant kitchen commands.


morgante:
Can I ask the motivation for this change?

aaron-lane:
@morgante `--test-base-path` is only valid for the commands contained in the first case. This change ensures we can use `kitchen_do` to run all `kitchen` commands, and hard-coded options like `--test-base-path` will only be applied when necessary.

morgante:
@aaron-lane Got it, thanks!

PR #374, opened 2019-10-11
Add parallel processing to scorecard.

briantkennedy:
Adds parallel decode JSON from reader with parallel calls to validator.Review.  This sets up a task graph in pipeline.go as follows:

read bytes -> unmarshal JSON & convert to validator.ReviewRequest -> call validator.Review -> results


morgante:
@briantkennedy Apologies for not reviewing this earlier.

@katze120 Have we had any issues with scalability in the field? If it proves to be an issue, we should look at this parallel processing. @yukinying This might help for ABCloud?

katze120:
Some benchmark data I recently collected from the field:

CAI file size | # of Projects | # of lines in CAI exports | Time to complete scorecard run with 55 constraints
-- | -- | -- | --
17M | 50 | 11567 | 4 mins
30M | 67 | 15508 | 6 mins
200M | 39047 | 247126 | 110 mins

Performance improvements would be nice to have. But would parallel processing stop us from being able to handle referential data in future?

morgante:
> Performance improvements would be nice to have. But would parallel processing stop us from being able to handle referential data in future?

It certainly might complicate it.

PR #385, opened 2019-10-17
Ocsig gke shared vpc 7

ocsig:


ocsig:
@sjvanrossum untested.

PR #408, opened 2019-10-28
add test/source.sh to set of ignored files

ideasculptor:


morgante:
Looks like commits are still being signed with your old email. You might have to amend/rebase them.

FYI you can easily see the email being used by appending ".patch" to the URL, https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/408.patch

ideasculptor:
@googlebot I signed it!

PR #411, opened 2019-10-28
Update folder.py.schema

frots:
Changed `rules` to `folders` in output reference


ocsig:
Thank you for your contribution!

PR #415, opened 2019-10-30
[developer-tools] Automatically pipe terraform outputs from setup folder into kitchen_do

### LINKED ISSUE #386: Getting rid of glue code which pipes tf outputs from setup to kitchen_do
I have a proposition to remove the requirement for explicitly defining the `test/setup/make_source.sh` script. Instead I think there is a pretty generic way to hide this functionality inside the `kitchen_do` command itself.

Here is what I think `kitchen_do` can perform right before the other stuff it does now:

1. run `terraform output -json` from the `test/setup` directory.
2. parse the JSON output, and dynamically do bash `export TF_VAR_{name}={value}` for each output value.
3. Handle the magic var `SERVICE_ACCOUNT_JSON` in a special way. I looked through the source code of several repos and looks like everywhere the naming and the command to obtain it is consistent. If the terraform variable `sa_key` exists, we'll perform the magic routine to setup the `SERVICE_ACCOUNT_JSON` out of it.
4. And we KEEP the old functionality -- if the `make_source.sh` exists, the setup script will execute it like it does now and the `kitchen_do` will just use it to override the variables defined implicitly. This way if there are some other magic variables like `SERVICE_ACCOUNT_JSON`, devs can just explicitly set them up.

I should be able to make prototype of this and integrate into several modules to test it pretty quickly. Since the dev tools live in a docker image, I can use python to parse json, so it should be pretty simple and straightforward to implement.

Would love to hear what you think about it.

cray0000:
- fix #386 

Implement a transport bus to automatically pipe terraform outputs from `test/setup` folder into `kitchen` tests as `TF_VAR_*` inputs.

NON-breaking, since the old way of doing it through `make_source.sh` is still supported. So it should be safe to use it without changing any code at all in the end user libraries.

Tested it using `iam` module as an example.

Supports the following usecases (tested different variations in `iam` module):

1. no extra files at all (just remove `make_source.sh`):
  - `docker_test_prepare` does nothing
  - in `docker_run` when executing `kitchen_do` it automatically retrieves outputs from `test/setup` automatically.
2. `make_source.sh` exists:
  - `docker_test_prepare` runs it and generates `test/source.sh` (same as before).
  - in `docker_run` when executing `kitchen_do` it automatically retrieves outputs from `test/setup` AND after that executes `test/setup.sh` (if it exists) which will add additional exports or overwrite some of the existing ones (if the name is the same).

In both cases if the `test/setup` folder doesn't exist or its terraform configuration is incorrect or wasn't provisioned, we only show it as a warning and keep executing the next command.


cray0000:
@googlebot I signed it!

morgante:
Thanks, released as v0.5.0.

PR #418, opened 2019-11-01
Accept arbitrary content before the license header

### LINKED ISSUE #390: verify_boilerplate.py should support accept arbitrary content before the license header
As seen in https://github.com/terraform-google-modules/terraform-google-container-vm/pull/43#pullrequestreview-304188817, there are cases where files are required to have content before the license header. `verify_boilerplate.py` should be updated to support this use case.

omazin:
Fixes #390


morgante:
@omazin Please make sure the test function also works/passes.

Issue #382, opened 2019-10-17
CLI reference within a value


ocsig:
CFT CLI should be able to handle references within a string such as:
`      - serviceAccount:service-$(out.myproject.myresource.myoutput.projectNumber)@container-engine-robot.iam.gserviceaccount.com`

PR #419, opened 2019-11-01
#382: CLI reference within a value

nick4fake:
Fixes https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/382

Added support for references in strings.


cli/deployment/config_test.go
```diff
@@ -69,8 +69,14 @@ func TestFindAllDependencies(t *testing.T) {
 		Project: "projectB",
 		Name:    "deploymentB",
 	}
+	configC := Config{
```
kopachevsky: need to update TestYamlReplaceOutRefs also, it's more relevant case

PR #447, opened 2019-11-13
Add subnetwork properties purpose & role

frots:
To allow setup of INTERNAL_HTTPS_LOADBALANCING reserved range subnet

### DIFF
--- a/dm/templates/network/subnetwork.py
+++ b/dm/templates/network/subnetwork.py
@@ -23,6 +23,8 @@ def generate_config(context):
     optional_properties = [
         'project',
         'enableFlowLogs',
+        'purpose',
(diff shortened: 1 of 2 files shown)


ocsig:
Something is wrong with the Cloudbuild trigger, but tested this PR manualy.
LGTM

frots:
Thanks

frots:
@ocsig -- when will this end up in master?

ocsig:
Appologies, I did not realized this is not targetting Master. Can you resend it to target master? The `cft-dm-dev` is not in a shape for merging. ( After that it will be in master within hours.) If you don't have the time, I will try to find some tonight.

PR #453, opened 2019-11-18
Changed GCS bucket 'setIamPolicy' action to virtual type

paulpalamarchuk:
* Changed action to the new virtual type
 * Added explicit dependencies to the GCS bindings
 * Added bats-test for 35 GCS bindings


paulpalamarchuk:
@kopachevsky Changed only internal logic. All examples are the same

PR #454, opened 2019-11-18
Add project labels check to the project.py.schema

### LINKED ISSUE #174: project labels regexp check
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/9c3192193e9706bdf74dad0482e9d87c93886d8d/dm/templates/project/project.py.schema#L180

> Keys and values can contain only lowercase letters, numeric characters, underscores, and dashes. All characters must use UTF-8 encoding, and international characters are allowed.
[labels](https://cloud.google.com/resource-manager/docs/creating-managing-labels)

paulpalamarchuk:
* Fix #174

### DIFF
--- a/dm/templates/project/project.py.schema
+++ b/dm/templates/project/project.py.schema
@@ -179,12 +179,24 @@ properties:
         pattern: ^[0-9]{8,25}$
   labels:
     type: object
+    minProperties: 0
+    maxProperties: 64
+    additionalProperties: false
+    patternProperties: 
+      ^[a-z](?:[-_a-z0-9]{0,61}[a-z0-9])?$ :
+        type: string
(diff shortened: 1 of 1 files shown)


dm/templates/project/project.py.schema
```diff
@@ -179,12 +179,24 @@ properties:
         pattern: ^[0-9]{8,25}$
   labels:
     type: object
+    minProperties: 0
```
kopachevsky: I don't sure about this one, how it correlates with doc:

**Keys have a minimum length of 1 character and a maximum length of 63 characters, and cannot be empty. Values can be empty, and have a maximum length of 63 characters.**

paulpalamarchuk: That correlates with:
**- Each resource can have multiple labels, up to a maximum of 64.**

PR #463, opened 2019-11-21
Kubernetes unified monitoring

trevorjwilliams:
Added support for what is referred to as "Enable Stackdriver Kubernetes Engine Monitoring" in the web console, and added constraints requiring the new setting to handle both logging and monitoring or neither for consistency with the web console. Also indicated that the older settings were "legacy" (again per the web console.)

Also added comments to a few fields to indicate that they were deprecated.


ocsig:
Please update the version in schema with a minor version. Add an entry to the changelog.

trevorjwilliams:
> Please update the version in schema with a minor version. Add an entry to the changelog.

Done.

PR #468, opened 2019-11-22
DM Validator - IAM member binding - Fixes 423

### LINKED ISSUE #423: DM Validator support Project, Folder, Org level IAM member binding
Depends on: #421 

Supporting the following GCP Types:
```
gcp-types/cloudresourcemanager-v1:virtual.organizations.iamMemberBinding
gcp-types/cloudresourcemanager-v2:virtual.folders.iamMemberBinding
gcp-types/cloudresourcemanager-v1:virtual.projects.iamMemberBinding
```

Note: This is not covering action based IAM setting.

nick4fake:
Fixes #423


cli/deployment/api_client.go
```diff
@@ -37,6 +37,29 @@ const (
 	ActionUpdate string = "update"
 )
 
+type DeploymentDescriptionResource struct {
```
kopachevsky: Сomment

cli/deployment/api_client.go
```diff
@@ -166,7 +190,7 @@ func CancelPreview(deployment *Deployment) (string, error) {
 		deployment.config.GetProject(),
 		"-q",
 	}
-	output, err := runGCloud(args...)
+	output, err := RunGCloud(args...)
```
kopachevsky: why it's public now?

cli/deployment/api_client.go
```diff
@@ -208,14 +232,40 @@ func Delete(deployment *Deployment, preview bool) (string, error) {
...
 	if err != nil {
 		log.Printf("failed to delete deployment: %v, error: %v, output: %s", deployment, err, output)
 		return output, err
 	}
 	return output, nil
 }
 
+func GetDeploymentDescription(name string, project string) (*DeploymentDescription, error) {
```
kopachevsky: comment

cli/deployment/api_client.go
nick4fake: Because I use it outside of the package

cli/deployment/api_client.go
nick4fake: Irrelevant to current PR

cli/validation/validate.go
```diff
@@ -36,26 +43,53 @@ func ValidateDeployment(name string, policyPath string, project string) (validat
...
+				asset.Ancestry, err = getAncestry(project)
+				if err != nil {
+					log.Printf("error getting resource ancestry: %s, deployment: %s, error: %v", res.Name, name, err)
+					return false, err
+				}
+
+				key := asset.Type + "__" + asset.Name
+
+				if existing, exists := assets[key]; exists {
+					// The existance of a merge function signals that this tf resource maps to a
```
morgante:
```suggestion
					// The existence of a merge function signals that this resource maps to a
```

cli/validation/validate.go
```diff
@@ -36,26 +43,53 @@ func ValidateDeployment(name string, policyPath string, project string) (validat
...
 
-			assets = append(assets, asset)
+				assets[key] = asset
+			}
 		} else {
 			log.Printf("resource type not supported: %s, name %s, deployment: %s, project %s\n", res.Type, res.Name, name, project)
 		}
 	}
 
+	getAssets := func(assets map[string]cai.Asset) byName {
```
morgante: It's pretty inefficient to rebuild this entire list at the end. Instead, I think we should be building the master slice from the beginning. We can, in parallel, also maintain a map() of names to asset indexes in the master list.

cli/validation/converter/iam.go
```diff
@@ -0,0 +1,90 @@
...
+				{
+					Role:    res.Properties["role"].(string),
+					Members: []string{res.Properties["member"].(string)},
+				},
+			},
+		},
+	}, nil
+}
+
+func MapIamCaiObject(existing cai.Asset, incoming cai.Asset) (cai.Asset, error) {
```
morgante: We should really call this `mergeIamAssets`. Calling it a map function is confusing.

cli/validation/converter/iam.go
```diff
@@ -0,0 +1,90 @@
+package converter
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/cli/validation/cai"
+)
+
+// mergeBindings clobbers members to bindings with the same roles
```
morgante: @ocsig Is this the right behavior in all cases?

Note that in Terraform we have two different merge functions: https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/master/google/iam_helpers.go#L72

The *logic* here is more or less a copy of `mergeAdditiveBindings`, while the comment is reflective of `mergeAuthoritativeBindings`.

Looking at the logic below, I *think* we're likely looking for the "additive" option (ie. not clobbering existing bindings for the same roles).

cli/validation/converter/iam_test.go
```diff
@@ -0,0 +1,330 @@
+package converter
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/cli/validation/cai"
+)
+
+func TestGetIamCaiObject(t *testing.T) {
```
morgante: This is fine for now, but I'd like to avoid handwritten tests for each resource. Instead we should have cases similar to what we do for TFV.

cli/validation/converter/iam_test.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+
+			if !reflect.DeepEqual(asset, c.output) {
+				t.Errorf("got %v, expected %v", asset.IAMPolicy, c.output.IAMPolicy)
+			}
+		})
+	}
+}
+
+func TestMergeBindings(t *testing.T) {
```
morgante: Any reason to duplicate this? https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/master/google/iam_helpers_test.go#L8

If we're copying the logic, we should probably copy the cases.

cli/validation/converter/iam.go
ocsig: Yes, we need the `mergeAdditiveBindings` functionality which this implementation does. The comment was incorrect, forgot to flag it.

cli/validation/converter/iam_test.go
nick4fake: I've added cases that haven't been covered by the conversion library.
cc @sjvanrossum

morgante: Then just add those cases on top of the existing ones. There's no reason to change the structure/logic.

In general, that's my biggest feedback: I don't want to see code drift when we're doing essentially the same thing.

nick4fake: It makes sense to me to add new tests to cover different function and set of logic. That's how unit tests work.

Do you want all unit tests to be in a single test case?

cli/validation/converter/iam_test.go
nick4fake: How do we do it for TFV?

morgante: https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/test/cli_test.go#L52

ocsig:
@morgante Can we merge this in?

PR #469, opened 2019-11-22
Enabled container API instead of compute

### LINKED ISSUE #433: Enabling container API instead of compute
When sharedVPC will be used for GKE it needs the container API and it's SA turned on.

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/859a8d014cd0b87ab5a62544053ed4d71d90ad12/dm/templates/project/project.py#L154

paulpalamarchuk:
* Fixed #433
 * Fixed GKE SA role binding dependency
 * Updated schema

### DIFF
--- a/dm/templates/project/project.py
+++ b/dm/templates/project/project.py
@@ -151,7 +151,7 @@ def activate_apis(context):
             properties.get('enableGKEToUseSharedVPC') and
             properties.get('sharedVPC')
         ):
-            apis.append('compute.googleapis.com')
+            apis.append('container.googleapis.com')
 
     resources = []
     api_names_list = ['{}-billing'.format(context.env['name'])]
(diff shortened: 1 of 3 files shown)


dm/templates/project/project.py
```diff
@@ -258,14 +258,14 @@ def create_service_accounts(context, project_id):
     network_list = [
         'serviceAccount:$(ref.{}-project.projectNumber)@cloudservices.gserviceaccount.com'.format(context.env['name'])
     ]
+    service_account_dep = ["{}-api-compute.googleapis.com".format(context.env['name'])]
```
ocsig: What if Compute API is not turned on? Shouldn't this be `service_account_dep = []`?

paulpalamarchuk: Reworked

PR #478, opened 2019-11-27
Remove outdated warning and correct the launch date 

laurenzlong:
"Private by default" has actually been delayed until Jan 15. I also removed the conflicting note.


ocsig:
Thanks for your contribution!

PR #479, opened 2019-11-27
allow scorecard output to csv and json format

katze120:


cli/scorecard/cmd.go
```diff
@@ -21,6 +23,9 @@ func init() {
 	Cmd.Flags().StringVar(&flags.policyPath, "policy-path", "", "Path to directory containing validation policies")
 	Cmd.MarkFlagRequired("policy-path")
 
+	Cmd.Flags().StringVar(&flags.outputPath, "output-path", "", "Path to directory to contain scorecard outputs. Output to console if not specified")
+	Cmd.Flags().StringVar(&flags.outputFormat, "output-format", "", "Format of scorecard outputs, can be txt, json or csv, default is txt")
```
morgante: Could we handle this flag using a viper flag + default? https://github.com/spf13/viper#establishing-defaults

cli/scorecard/score.go
```diff
@@ -148,19 +150,67 @@ func (inventory *InventoryConfig) Score(config *ScoringConfig) error {
...
-			fmt.Printf("\n\n%v: %v issues found\n", category.Name, category.Count())
-			fmt.Printf("----------\n")
-			for _, cv := range category.constraints {
-				fmt.Printf("%v: %v issues\n", cv.GetName(), cv.Count())
-				for _, v := range cv.Violations {
-					fmt.Printf("- %v\n\n",
-						v.Message,
-					)
-					Log.Debug("Violation metadata", "metadata", v.GetMetadata())
+		content := ""
```
morgante: Instead of building a string in-memory + then writing it to the chosen destination (which is potentially inefficient), can we *first* build an io.Writer for the destination then have the various outputters write to it directly?

cli/scorecard/score.go
```diff
@@ -148,19 +150,67 @@ func (inventory *InventoryConfig) Score(config *ScoringConfig) error {
...
+		switch outputFormat {
+		case "json":
+			type violationOutput struct {
```
morgante: Let's not define a type inline here. Instead maybe we can define a new `RichViolation` struct? We should also have a constructor for said struct which can create it from a violation + category.

cli/scorecard/score.go
```diff
@@ -148,19 +150,67 @@ func (inventory *InventoryConfig) Score(config *ScoringConfig) error {
...
+						vOutput.Category = category.Name
+						vOutput.Constraint = v.Constraint
+						vOutput.Resource = v.Resource
+						vOutput.Message = v.Message
+						byteContent, err := json.MarshalIndent(vOutput, "", "  ")
+						if err != nil {
+							return err
+						}
+						content += string(byteContent) + "\n"
+						Log.Debug("Violation metadata", "metadata", v.GetMetadata())
```
morgante: Shouldn't we simply include metadata in the JSON output?

cli/scorecard/score.go
```diff
@@ -148,19 +150,67 @@ func (inventory *InventoryConfig) Score(config *ScoringConfig) error {
...
+					}
+				}
+			}
+		case "csv":
+			content = "Category,Constraint,Resource,Message\n"
+			for _, category := range config.categories {
+				for _, cv := range category.constraints {
+					for _, v := range cv.Violations {
+						content += fmt.Sprintf("%v,%v,%v,%v\n", category.Name, v.Constraint, v.Resource, v.Message)
```
morgante: I think we should use Golang's CSV writer instead of trying to manually construct ourselves.

We could define a `ToCSV` function on the `RichViolation` type which can create a slice for easy output. https://stackoverflow.com/questions/43301956/convert-struct-to-slice-of-strings

cli/scorecard/cmd.go
katze120: addressed in e4926b7

cli/scorecard/score.go
katze120: addressed in 6c16093

cli/scorecard/score.go
katze120: addressed in 372a085

cli/scorecard/score.go
katze120: included in 372a085

cli/scorecard/score.go
katze120: used csv writer in 18e499c
but RichViolation is not used for csv output because metadata is in the struct, I'd prefer csv to include simply string fields only and it's easier to just call those fields.

katze120:
thanks for the review feedback !

PR #483, opened 2019-11-27
Add a configurable exclude directory option for linting

milesmatthias:


infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -57,6 +57,7 @@ maketemp() {
 # find_files is a helper to exclude .git directories and match only regular
 # files to avoid double-processing symlinks.
 find_files() {
+  EXCLUDE_LINT_DIR=${EXCLUDE_LINT_DIR:-}
```
morgante: @milesmatthias Instead of only allowing a single directory, could we accept a list of directories?

morgante:
@milesmatthias Will you have a chance to try that?

milesmatthias:
> @milesmatthias Will you have a chance to try that?

I most likely won't have a chance to try that, so if you'd like to give that task to someone else, I'd suggest that.

PR #489, opened 2019-11-29
Fixes #488: DM Validator functional tests

### LINKED ISSUE #488: DM Validator functional tests
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/456

Splitting from unit tests as we need to rewrite bats tests to Go.

nick4fake:
Fixes https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/488

Added functional tests.

### DIFF
--- a/cli/cmd/common.go
+++ b/cli/cmd/common.go
@@ -23,6 +23,7 @@ var policyPathFlag string
 var validateFlag bool = false
 var previewFlag bool = false
 var showStagesFlag bool = false
+var nonInteractiveFlag bool = false
 var formatFlag string
 
 var supportedExt = []string{"*.yaml", "*.yml", "*.jinja"}
@@ -126,7 +127,10 @@ func executeStages(action string, stages [][]deployment.Config, isDelete bool) m
 			}
 
 			if previewFlag {
-				choice := deployment.GetUserInput("Update(u), Skip (s), or Abort(a) Deployment?", []string{"u", "s", "a"}, os.Stdin)
+				choice := "a"
+				if !nonInteractiveFlag {
+					choice = deployment.GetUserInput("Update(u), Skip (s), or Abort(a) Deployment?", []string{"u", "s", "a"}, os.Stdin)
+				}
 				switch choice {
 				case "u":
 					output, err := deployment.ApplyPreview(dep)
@@ -137,7 +141,7 @@ func executeStages(action string, stages [][]deployment.Config, isDelete bool) m
 				case "s":
 					_, err = deployment.CancelPreview(dep)
 					if err != nil {
-						log.Fatalf("error cancel-preuvew action for deployment: %v, error: %v", dep, err)
+						log.Fatalf("error cancel-preview action for deployment: %v, error: %v", dep, err)
 					}
 					log.Printf("canceled %s action for deployment: %v", action, dep)
 					if action == deployment.ActionCreate {
--- a/cli/tests/cli_test.go
+++ b/cli/tests/cli_test.go
@@ -1,0 +1,245 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+var (
+	cwd       string
+	cftBinary string
+)
+
+// TestCLI does integration testing.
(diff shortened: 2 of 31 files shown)


cli/cmd/common.go
```diff
@@ -23,10 +23,15 @@ var policyPathFlag string
 var validateFlag bool = false
 var previewFlag bool = false
 var showStagesFlag bool = false
+var nonInteractiveFlag bool = false
 var formatFlag string
 
 var supportedExt = []string{"*.yaml", "*.yml", "*.jinja"}
 
+func initNonInteractiveFlag(command *cobra.Command) {
```
morgante: We should simply add this on the root cli command (it should be consistent across *all* cli commands)

cli/tests/cli_test.go
```diff
@@ -0,0 +1,245 @@
...
+	"strings"
+	"testing"
+)
+
+var (
+	cwd       string
+	cftBinary string
+)
+
+// TestCLI does integration testing.
```
morgante: I think this function is attempting to be overly generic. We can have a testing function with cases focused solely on testing validation functionality.

If there is useful code which we can share with other integration tests, we could/would put it into shared functions.

cli/tests/cli_test.go
```diff
@@ -0,0 +1,245 @@
...
+	}{
+		{
+			name: "validate",
+			testers: []struct {
+				name   string
+				tester func(t *testing.T, dirTestdata string)
+			}{
+				{
+					name:   "passing",
+					tester: testValidatePassing,
```
morgante: As mentioned, these test cases and functions seem a bit overly generic. I'd try to make this much closer to the tf-validator tests.

In fact, we can probably mirror those almost exactly - https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/test/cli_test.go#L49

cli/tests/cli_test.go
```diff
@@ -0,0 +1,245 @@
...
+	})
+	checkOutput(t, append(stdOut, stdErr...), []string{
+		"Found Violations",
+		"Constraint iam_ban_roles",
+		"//cloudresourcemanager\\.googleapis\\.com/projects/12345",
+	})
+}
+
+func setupValidate(t *testing.T, dirTestdata string) {
+	runCft(t, false, cwd, []string{"apply", filepath.Join(dirTestdata, "deployment")})
```
morgante: Validation tests should ideally not require creating *any* actual live resources. (It doesn't for TFV.)

cli/tests/cli_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+func runCft(t *testing.T, wantError bool, dir string, args []string) ([]byte, []byte) {
+	executable := cftBinary
+	args = append(args, "--non-interactive")
+	cmd := exec.Command(executable, args...)
+	cmd.Dir = dir
+	return run(t, cmd, wantError)
+}
+
+// run a command and call t.Fatal on non-zero exit.
```
morgante: Don't copy & paste from TFV, just import - https://github.com/GoogleCloudPlatform/terraform-validator/blob/master/test/cli_test.go#L253

cli/tests/cli_test.go
```diff
@@ -0,0 +1,245 @@
...
+		t.Log(stdout.String())
+	}
+	if stderr.String() != "" {
+		t.Logf("=== STDERR of %s ===", cmdToString(cmd))
+		t.Log(stderr.String())
+	}
+	return stdout.Bytes(), stderr.Bytes()
+}
+
+// cmdToString clones the logic of https://golang.org/pkg/os/exec/#Cmd.String.
```
morgante: Shouldn't need this at all.

cli/tests/validate/policies/test_none/policies/templates/gcp_iam_allow_ban_roles_v1.yaml
```diff
@@ -0,0 +1,110 @@
+# Copyright 2019 Google Inc.
```
morgante: I don't want to proliferate this repo with lots of resource-specific tests. We can stick to using always_violates or the banned_resource test for basically everything.

cli/tests/cli_test.go
nick4fake: There has been no existing tests implementation. I've created it being inspired by terraform validator tests.

I haven't moved code that could be shared to separate functions because I've been ignoring other tests completely (as you and @aaron-lane have asked me to do)

cli/tests/cli_test.go
nick4fake: No really, because logic and lifecycle is different.

For example, DM requires real resources to be created (even in preview mode we need to create previews)

cli/tests/cli_test.go
nick4fake: I don't think it is possible. Terraform is different from DM

cli/tests/cli_test.go
nick4fake: How can I import unexported function?

cli/tests/cli_test.go
nick4fake: Why? It is used for logging purposes.

cli/tests/validate/policies/test_none/policies/templates/gcp_iam_allow_ban_roles_v1.yaml
nick4fake: The main reason I'm adding resource specific tests is to check that properties are correctly mapped.

PR #496, opened 2019-12-03
Added log sink template with external bucket

### LINKED ISSUE #77: Log sink template could be more flexible
Our current deployment looks something like this:
- Log sink deployed on project-foo
- Get log sink writer and set in dataset
- Deploy dataset in project-bar

Two deployments in two projects.

The current template has a few problems with this:
- it doesn't allow cross project sinks
- due to the writer service account being in the form seviceAccount:foo@google.com the bigquery access field cannot be set.

It would be great if cross project sinks + some workaround for the dataset could be made. Perhaps it can do some project id matching inside the template and the same template can be deployed in both projects and it will be smart enough to know which resource to deploy in which project?

kopachevsky:
* Fix #77

### DIFF
--- a/dm/templates/logsink/logsink.py
+++ b/dm/templates/logsink/logsink.py
@@ -121,7 +121,7 @@ def create_storage(context, logsink_name):
         dest_prop['project'] = project_id
         bindings = dest_prop.get('bindings', [])
         bindings.append({
-            'role': 'roles/storage.objectAdmin',
+            'role': 'roles/storage.admin',
             'members': ['$(ref.{}.writerIdentity)'.format(logsink_name)]
         })
 
--- a/dm/templates/logsink/logsink.py
+++ b/dm/templates/logsink/logsink.py
@@ -18,7 +18,7 @@ def create_pubsub(context, logsink_name):
     """ Create the pubsub destination. """
 
     properties = context.properties
-    project_id = properties.get('project', context.env['project'])
+    project_id = properties.get('destinationProject', properties.get('project', context.env['project']))
 
     dest_properties = []
     if 'pubsubProperties' in context.properties:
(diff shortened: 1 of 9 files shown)


dm/templates/logsink/logsink.py
```diff
@@ -19,6 +19,7 @@ def create_pubsub(context, logsink_name):
 
     properties = context.properties
     project_id = properties.get('project', context.env['project'])
```
ocsig: Instead of the 2 line to set and overwrite project_id, I suggest a single line:
```suggestion
    project_id = properties.get('destinationProject', properties.get('project', context.env['project']))
```

Please update all places

dm/templates/logsink/logsink.py
```diff
@@ -19,6 +19,7 @@ def create_pubsub(context, logsink_name):
...
+    project_id = properties.get('destinationProject', project_id)
```
ocsig: Instead of the 2 line to set and overwrite project_id, I suggest a single line:
```suggestion
    project_id = properties.get('destinationProject', properties.get('project', context.env['project']))
```

Please update all places

dm/templates/logsink/logsink.py
```diff
@@ -89,39 +123,50 @@ def create_storage(context, logsink_name):
...
+                        }]
+                    }
+            }
+        ]
+
+        if len(bindings) > 0:
+            dest_properties.append({
+                'name': '{}-iampolicy'.format(bucket_name),
                 # https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy
                 'action': 'gcp-types/storage-v1:storage.buckets.setIamPolicy',
```
ocsig: Why are we not using the virtual IAMMemberBinding type?

ocsig: Or the iam_member binding

dm/templates/logsink/logsink.py
```diff
@@ -89,36 +120,32 @@ def create_storage(context, logsink_name):
...
         dest_prop['project'] = project_id
         bindings = dest_prop.get('bindings', [])
-        bindings.append(
-            {
-                'role': 'roles/storage.admin',
-                'members': ['$(ref.' + logsink_name + '.writerIdentity)']
-            }
-        )
+        bindings.append({
+            'role': 'roles/storage.objectAdmin',
```
ocsig: This is breaking out bats test!

ocsig: in test file templates/logsink/tests/integration/logsink.bats, line 169)
#   `[[ "$output" =~ "roles/storage.admin" ]]' failed

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/dm/templates/logsink/tests/integration/logsink.bats

PR #503, opened 2019-12-06
Added support of bucketPolicyOnly setting for gcs_bucket template

### LINKED ISSUE #455: bucketPolicyOnly setting isn't working for gcs_bucket template
When I attempt to use `gcs_bucket.py` to create a bucket with `iamConfiguration.bucketPolicyOnly.enabled: True`, I get the error message:

```Cannot use ACL API to set bucket policy when object policies are disabled, must use IAM API instead.```

I'm able to work around this by removing the `default` settings from the `predefinedAcl` and `predefinedDefaultObjectAcl` values in `gcs_bucket.py.schema`.

### PR DESCRIPTION
**Pull request by kopachevsky**
* Fix #455

### DIFF
--- a/dm/templates/gcs_bucket/gcs_bucket.py
+++ b/dm/templates/gcs_bucket/gcs_bucket.py
@@ -66,12 +66,18 @@ def generate_config(context):
         if prop in properties:
             bucket['properties'][prop] = properties[prop]
 
+    if not properties.get('iamConfiguration', {}).get('bucketPolicyOnly', {}).get('enabled', False):
+        if 'predefinedAcl' not in bucket['properties']:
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: dm/templates/gcs_bucket/gcs_bucket.py [hunk 1]
```diff
@@ -66,12 +66,18 @@ def generate_config(context):
         if prop in properties:
             bucket['properties'][prop] = properties[prop]
 
+    if properties.get('iamConfiguration', {}).get('bucketPolicyOnly', {}).get('enabled', False) is not True:
```
ocsig:
```suggestion
    if not properties.get('iamConfiguration', {}).get('bucketPolicyOnly', {}).get('enabled', False):
```

PR #509, opened 2019-12-09
Add varable to set additional dirs to exclude from lints

paulpalamarchuk:
* Based on PR 483
 * ref https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/483#pullrequestreview-323981445


milesmatthias:
@paulpalamarchuk was this tested with wildcards & subdirectories?

I was able to get a root level directory, but not subdirectories. See PRs:

https://github.com/terraform-google-modules/terraform-google-bigquery/pull/38
https://github.com/terraform-google-modules/terraform-google-bigquery/pull/35

paulpalamarchuk:
@milesmatthias please use relative path. Example:
```
("\./examples" "\./example/example\.py")
```

milesmatthias:
Hmm... that's not working for me:

```
.PHONY: docker_test_lint
docker_test_lint:
	docker run --rm -it \
		-e "EXCLUDE_LINT_DIRS":'("\./modules/view_generator/scripts")' \
		-v $(CURDIR):/workspace \
		$(REGISTRY_URL)/${DOCKER_IMAGE_DEVELOPER_TOOLS}:${DOCKER_TAG_VERSION_DEVELOPER_TOOLS} \
		/usr/local/bin/test_lint.sh
```

I've also tried escaping the underscore...

milesmatthias:
So the rub here is that environment variables need to be simple key value pairs, not arrays. I should've caught that earlier, sorry.

Here's an updated PR that leverages your looping code but takes a string rather than an array for user-supplied dirs to ignore in linting: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/523

paulpalamarchuk:
@milesmatthias it's not working, you are right. I haven't tested passing env var from outside to the docker. Docker doesn't accept array env vars.

PR #510, opened 2019-12-09
[test-org] Create a secrets bucket to store gsuite sa key in.

cray0000:
Create a new bucket to store and later retrieve the gsuite key.

Also has an example on how to retrieve the key as the comment.

```hcl
data "google_storage_object_signed_url" "ci_gsuite_sa_json" {
  bucket   = "ci-gsuite-sa-secrets"
  path     = "gsuite-sa.json"
  duration = "1m"
}
data "http" "ci_gsuite_sa_json" {
  url = data.google_storage_object_signed_url.ci_gsuite_sa_json.signed_url
}
output "gsuite_sa_json" {
  value     = data.http.ci_gsuite_sa_json.body
  sensitive = true
}
```

### DIFF
--- a/infra/terraform/test-org/org/gsuite.tf
+++ b/infra/terraform/test-org/org/gsuite.tf
@@ -109,14 +109,7 @@ resource "google_service_account_key" "ci_gsuite_sa" {
   service_account_id = google_service_account.ci_gsuite_sa.id
 }
 
-provider "google" {
-  alias       = "sa"
-  credentials = base64decode(google_service_account_key.ci_gsuite_sa.private_key)
-}
-
 resource "google_storage_bucket" "ci_gsuite_sa" {
-  provider = google.sa
-
   name          = local.ci_gsuite_sa_bucket
   storage_class = "MULTI_REGIONAL"
   project       = module.ci_gsuite_sa_project.project_id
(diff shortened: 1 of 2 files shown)


infra/terraform/test-org/org/gsuite.tf
```diff
@@ -99,8 +103,61 @@ resource "google_billing_account_iam_member" "ci_gsuite_sa_billing" {
...
+resource "google_service_account_key" "ci_gsuite_sa" {
+  service_account_id = google_service_account.ci_gsuite_sa.id
+}
+
+provider "google" {
+  alias       = "sa"
+  credentials = base64decode(google_service_account_key.ci_gsuite_sa.private_key)
+}
+
+resource "google_storage_bucket" "ci_gsuite_sa" {
```
morgante: Why does this bucket need to be created with a different provider? The normal/default one should work fine.

infra/terraform/test-org/org/gsuite.tf
```diff
@@ -99,8 +103,61 @@ resource "google_billing_account_iam_member" "ci_gsuite_sa_billing" {
...
 
-# Grant G-Suite project rights to cft_ci_group
-# Required to be able to create keys for the gsuite sa.
+// Generate a json key and put it into the secrets bucket.
+
+resource "google_service_account_key" "ci_gsuite_sa" {
+  service_account_id = google_service_account.ci_gsuite_sa.id
+}
+
+provider "google" {
```
morgante: I don't see any need for a separate provider config.

cray0000: Yep, thanks. I was testing it locally launching from the service account, trying to store and retrieve it, and mixed it up. We have to use a custom provider config launched from SA and not personal, to retrieve the key, since signed url can be created only under SA.

Fixed and updated the retrieval example in the PR description.

PR #514, opened 2019-12-12
Limited ports to use for the TCP external proxy according to docs

### LINKED ISSUE #494: [DM] Limit ports that can be used for TCP Proxy LB
As of `TCP Proxy Load Balancing supports the following ports: 25, 43, 110, 143, 195, 443, 465, 587, 700, 993, 995, 1883, 5222`, TCP-proxy ports should be limited. That related to the `target_proxy.py` template and all templates that use it.

paulpalamarchuk:
* Fix #494


ocsig:
Test is correct, cleanup has known issue.

PR #515, opened 2019-12-12
Fixed bug in template name value

### LINKED ISSUE #502: Bug in firewall template
When  `sha1(rule['name']).hexdigest()[:10]` generates a hash that matches an exponential number notation, DM errors with something like:
```googlecloudsdk.api_lib.deployment_manager.exceptions.OperationError: Error in Operation [operation-000000-00000-000-000]: errors:
- code: INVALID_CONFIG
  message: "Invalid config files: Template fetching failed: Failed to parse template\
    \ name 7.7727716E8 since it's not a String.."
```

In this case the hash function resulted in string '77727716e8' and apparently somewhere it is cast to non string type.

kopachevsky:
* Fix #502


morgante:
@ocsig Can we merge this change?

PR #521, opened 2019-12-19
Allow CFT Scorecard to read JSON metadata from Stdin

pwilthew:
Update to allow CFT Scorecard to read IAM and Resources inventories from Stdin when bucket and dir-path where not passed.


pwilthew:
Added the --stdin boolean flag.

PR #522, opened 2019-12-20
allow passing in validator directly to construct ScoringConfig

yukinying:


yukinying:
@morgante can you PTAL? Thanks.

yukinying:
> Could you remove it and all references?

Done. The field is removed, together with the reference of it (`attachValidator`).

PR #523, opened 2019-12-22
fix ability to add multiple directories to ignore in linting

milesmatthias:
Also added comments to make it clear what's going on here and what should be expected, which should make it easier for people to see how to use this.

Fixes https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/509


milesmatthias:
@googlebot add-reviewer @paulpalamarchuk

Issue #529, opened 2020-01-02
GKE template is incompatible with Python3


buhkguj:
Hi,
The dm/templates/gke/ template produces the following warning in DM newly:

> The template is incompatible with Python3. Please fix the following errors: Error compiling Python code: No module named 'packaging' Resource: gke.py Resource: config https://cloud.google.com/deployment-manager/docs/migrate-to-python3

patrickmslatteryvt:
Yes, I'm getting that too, nothing I have tried fixes it.

trevorjwilliams:
I believe it's only a warning, and so can be ignored for the moment.

However, it's caused by logic that looks to exist only to help with versions of kubernetes older than 1.12, which are no longer options from the web console at least. Consequently, you can get rid of the warning by deleting the corresponding logic, [as i did here](https://github.com/trevorjwilliams/cloud-foundation-toolkit/commit/5abcd362f118c7721cf10ba22d038517df421b73).

Not sure this is the preferred solution, however.

morgante:
@trevorjwilliams Would you mind opening a PR with that fix? I agree we can drop support for Kubernetes versions older than 1.12.

trevorjwilliams:
Sure, I'll update the version number, add a comment, and submit it.

buhkguj:
PR fixed the warning. Thx!

PR #531, opened 2020-01-08
Update gke template to support python3

trevorjwilliams:
- Deleted logic supporting versions of Kubernetes older than 1.12, allowing removal of dependency on `packaging`.


trevorjwilliams:
Fixes #529

ocsig:
Thanks for your contribution!

PR #536, opened 2020-01-12
feat: add support for outputting drawing DSL from launchpad CLI

morgante:
This adds support for launchpad to output diagrams/drawings.

### DIFF
--- a/cli/launchpad/draw.go
+++ b/cli/launchpad/draw.go
@@ -1,0 +1,132 @@
+package launchpad
+
+import (
+	"fmt"
+	"io"
+	"strings"
+)
+
+// diagram represents a GCP draw diagram
+type diagram struct {
+	paths []*diagramPath
+	group *diagramGroup
+}
+
+// diagramGroup represents a group of cards (including a top-level diagram)
+type diagramGroup struct {
+	d      *diagram
+	name   string
+	cards  []*diagramCard
+	groups []*diagramGroup
+}
+
+// diagramPath represents a path within a GCP diagram
+type diagramPath struct {
+	from      *diagramCard
+	to        *diagramCard
+	connector string
+}
+
+// diagramCard represents a card within a GCP diagram
+type diagramCard struct {
+	id   string
+	name string
+}
+
+// addCard adds a card into a diagram
+func (dg *diagramGroup) addCard(id string, displayName string) (*diagramCard, error) {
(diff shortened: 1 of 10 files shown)


morgante:
@mikeweng Could you take a look?

cli/launchpad/draw.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"io"
+	"strings"
+	"os"
+)
+
+// diagram represents a GCP draw diagram
+type diagram struct {
+	paths []*diagramPath
+	group *diagramGroup
```
mikeweng: We should map this to the diagram syntax. 

Group essentially is a wrapper for a list of card and other nested groups. And as this code indicates, top level elements{} essentially equates to a group. Therefore, maybe a new name to avoid conflict 
Maybe something similar to this? 
```
type diagram struct {
  meta *diagramMeta
  elements *diagramElements
  paths []*diagramPath
}
type diagramElements diagramGroup
```

cli/launchpad/draw.go
```diff
@@ -0,0 +1,141 @@
...
+	name   string
+	cards  []*diagramCard
+	groups []*diagramGroup
+}
+
+// diagramPath represents a path within a GCP diagram
+type diagramPath struct {
+	from      *diagramCard
+	to        *diagramCard
+	connector string
```
mikeweng: should we do enum for all the different types of connectors

cli/launchpad/draw.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// diagramCard represents a card within a GCP diagram
+type diagramCard struct {
+	id   string
+	name string
```
mikeweng: does card include stacked_card?
also, we can attach description as well.

cli/launchpad/draw.go
```diff
@@ -0,0 +1,141 @@
...
+
+// diagram represents a GCP draw diagram
+type diagram struct {
+	paths []*diagramPath
+	group *diagramGroup
+}
+
+// diagramGroup represents a group of cards (including a top-level diagram)
+type diagramGroup struct {
+	d      *diagram
```
mikeweng: I'm wondering if this is needed, when does a group need access its diagram? I see only on creating another group under the same diagram. Is this to pass some metadata?

PR #537, opened 2020-01-13
Launchpad nested folders and conflict definitions

### LINKED ISSUE #535: Launchpad doesn't handle nested folders
CFT Launchpad doesn't seem to currently support nested folders.

This is the config used:
```
apiVersion: cft.dev/v1alpha1
kind: Organization
spec:
  # TODO: capture real org ID
  id: 12345678
  folders:
  - id: "dev"
    displayName: "dev"
    folders:
    - id: "shared"
      displayName: "shared"
  - id: "prod"
    displayName: "production"
    folders:
      - id: "shared"
        displayName: "shared"
```

Command: `cft lp generate CloudFoundation.yaml -d terraform`

Actual output:
```
2020/01/10 16:36:47 debug: output location terraform
2020/01/10 16:36:47 debug: output flavor Terraform
2020/01/10 16:36:47 1 YAML documents loaded
Resource Map [4]:
  * Folder.dev <- []
  * Folder.prod <- []
  * Folder.qa <- []
  * Organization.12345678 <- [Folder.dev, Folder.prod, Folder.qa]
Organization.12345678 ("")
  + Folder.dev ("dev") < Organization.12345678
  + Folder.prod ("production") < Organization.12345678
  + Folder.qa ("QA") < Organization.12345678
```

Note that the shared folders don't show up anywhere.

mikeweng:
Added missing code for nested folders and detecting conflicts on folder definitions.

Fixes #535.

### DIFF
--- a/cli/launchpad/folder.go
+++ b/cli/launchpad/folder.go
@@ -26,13 +26,27 @@ type folderSpecYAML struct { // Inner mappings
 type folders []*folderYAML
 
 // add appends a folder into the folder list if it does not exist already.
-func (fs *folders) add(newF *folderYAML) {
+//
+// add returns existing folder's reference if attempted to add folder of the same ID.
+func (fs *folders) add(newF *folderYAML) *folderYAML {
 	for _, f := range *fs {
-		if f.Spec.Id == newF.Spec.Id { // silently ignore already existing folder
-			return
+		if f.Spec.Id == newF.Spec.Id {
+			return f
 		}
 	}
 	*fs = append(*fs, newF)
+	return nil
+}
+
+// merge adds given list of folders into the current folders list recursively.
+//
+// merge will be called on subfolder when it exist in both current and given list of folders.
+func (fs *folders) merge(nfs *folders) {
+	for _, f := range *nfs {
+		if existF := fs.add(f); existF != nil {
+			existF.subFolders.merge(&f.subFolders) // recursively merge sub folders
+		}
+	}
 }
 
 func (fs folders) sortedCopy() folders {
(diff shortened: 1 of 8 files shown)


mikeweng:
Addressing #535

cli/launchpad/folder.go
```diff
@@ -26,13 +26,27 @@ type folderSpecYAML struct { // Inner mappings
...
 	*fs = append(*fs, newF)
+	return nil
+}
+
+// merge adds given list of folders into the current folders list recursively.
+//
+// merge will be called on subfolder when it exist in both current and given list of folders.
+func (fs *folders) merge(nfs *folders) {
+	for _, f := range *nfs {
+		if existF := fs.add(f); existF != nil {
```
morgante: Shouldn't we *always* add? I don't necessarily see why `merge` should only happen on existing folders.

mikeweng: Ah, this merge happens when we are joining two lists together. The only possible recursive call is when a folder exists in both list. ex:
`fs = []folder{f1, f2, f3}; nfs = []folder{f3, f4, f5}`
Expected output of `fs.merge(nfs)` would be `fs=[]folder{f1, f2, f3, f4, f5}`. 
Calling merge on `f3` is natural because `f3` in `fs` might have subfolder `f31, f32`, and `f3` in `nfs` might have `f32, f33`, which we expect eventually `f3` in `fs` to be `f31, f32, f33`. 
However, calling f1 to merge is unnecessary because `f1` doesn't exist in `nfs` and it's a unique copy. Similar on `f5` that could just be added onto `fs` without a merge operation.

PR #544, opened 2020-01-23
scorecard - Cloud Asset Inventory refresh feature

katze120:
Feature to refresh Cloud Asset Inventory export files in GCS bucket

### DIFF
--- a/cli/docs/scorecard.md
+++ b/cli/docs/scorecard.md
@@ -31,7 +31,7 @@ This tutorial will walk you through setting up Scorecard for a single project.
     gsutil mb gs://$PUBLIC_BUCKET_NAME
     gsutil iam ch allUsers:objectViewer gs://$PUBLIC_BUCKET_NAME
     ```
-6. Export the CAI data to GCS:
+6. Optionally, export the CAI data to GCS:
     ```
     # Export resource data
     gcloud asset export --output-path=gs://$CAI_BUCKET_NAME/resource_inventory.json \
@@ -44,6 +44,7 @@ This tutorial will walk you through setting up Scorecard for a single project.
         --project=$GOOGLE_PROJECT \
         # could also use --folder or --organization
     ```
+    The alternative is to use [integrated inventory refresh feature](#Using-integrated-inventory-refresh-feature) `cft scorecard --refresh`
 7. Download the CFT CLI and make it executable:
     ```
     # OS X
--- a/cli/docs/scorecard.md
+++ b/cli/docs/scorecard.md
@@ -26,7 +26,12 @@ This tutorial will walk you through setting up Scorecard for a single project.
     ```
     gsutil mb gs://$CAI_BUCKET_NAME
     ```
-5. Export the CAI data to GCS:
+5. Optionally, create a public GCS bucket to trigger a violation:
+    ```
+    gsutil mb gs://$PUBLIC_BUCKET_NAME
+    gsutil iam ch allUsers:objectViewer gs://$PUBLIC_BUCKET_NAME
+    ```
+6. Export the CAI data to GCS:
     ```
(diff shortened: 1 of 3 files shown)


cli/docs/scorecard.md
```diff
@@ -26,7 +26,12 @@ This tutorial will walk you through setting up Scorecard for a single project.
...
-5. Export the CAI data to GCS:
+5. Optionally, create a public GCS bucket to trigger a violation:
+    ```
+    gsutil mb gs://$PUBLIC_BUCKET_NAME
+    gsutil iam ch allUsers:objectViewer gs://$PUBLIC_BUCKET_NAME
+    ```
+6. Export the CAI data to GCS:
     ```
     # Export resource data
     gcloud asset export --output-path=gs://$CAI_BUCKET_NAME/resource_inventory.json \
```
morgante: If we're reintroducing refresh can we skip this step?

cli/scorecard/inventory.go
```diff
@@ -29,49 +29,65 @@ import (
 // InventoryConfig manages a CAI inventory
 type InventoryConfig struct {
 	projectID        string
-	controlProjectID string
+	folderID   		 string
 	organizationID   string
 	bucketName       string
 	dirPath          string
 	readFromStdin    bool
+	refresh			 bool
```
morgante: I don't think there's any need to store this on the config?

cli/scorecard/inventory.go
```diff
@@ -29,49 +29,65 @@ import (
...
+			return nil, err
+		}
+	}
 	return inventory, nil
 }
 
 func (inventory InventoryConfig) getParent() string {
 	if inventory.organizationID != "" {
 		return fmt.Sprintf("organizations/%v", inventory.organizationID)
+	}else if inventory.folderID != "" {
```
morgante:
```suggestion
	} else if inventory.folderID != "" {
```

cli/scorecard/inventory.go
```diff
@@ -130,11 +146,12 @@ func (inventory *InventoryConfig) Export() error {
 		return err
 	}
 	err = inventory.exportToGcs(assetpb.ContentType_IAM_POLICY)
-	s.Stop()
-
 	if err != nil {
+		s.Stop()
```
morgante: I'm not sure why this needs to be duplicated?

cli/docs/scorecard.md
katze120: Updated to make this step optional (c9be8ea)

Refresh feature is only functional if the user runs scorecard as a service account. But the purpose of this particular guide and even the positioning of scorecard is mostly for ad-hoc runs and quick evaluations, likely on the user's local environment, hence I'd prefer not encourage user to download a service account key and keep the `gcloud asset export` option here because the gcloud command works with end user credential.

cli/scorecard/inventory.go
katze120: removed in c317f93

cli/scorecard/inventory.go
katze120: removed in c317f93

cli/scorecard/inventory.go
katze120: removed in c317f93

cli/scorecard/inventory.go
katze120: added space in 6a9327a

cli/scorecard/inventory.go
katze120: It's to remove the spinning message from console when each individual task runs into error

cli/docs/scorecard.md
morgante: Got it, makes sense.

PR #547, opened 2020-01-28
scorecard - new feature to include additional metadata fields in output

katze120:
New scorecard feature to include violation metadata fields of string type in txt or csv output. 

Use cases include a user-defined priority for each constraint, resource parent/project number/project id, or any additional information that is not covered in resource name and violation message.

The expectation is that these metadata fields need to be defined in constraint templates.

### DIFF
--- a/cli/scorecard/cmd.go
+++ b/cli/scorecard/cmd.go
@@ -19,7 +19,7 @@ var flags struct {
 	refresh          bool
 	outputPath       string
 	outputFormat     string
-	metadata		 []string
+	metadataFields		 []string
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


katze120:
Regarding letting user list out every metadata field they want:

In csv format, this gives users full control on the columns in the output and sequence of columns, which makes it possible for further processing of the output via spreadsheet formulas, hence addressing some immediate use cases from the field, such as:
- referential checks
- dashboarding & filtering by project
- dashboarding & filtering by priority
Currently a workaround is composing the info as delimited string in violation message. But since metadata is available, leveraging metadata would be neater way to handle this

Regarding simply having a flag to turn all metadata on and off:

I thought of 2 possible approaches: output all metadata in json format in one single field, or output as expanded fields
Outputting all metadata in json format in one single field 1 makes txt output less human readable and csv output less possible for further processing via spreadsheet formula. For programmatic consumption of output, users can use continue to use json format.
Outputting all metadata as expanded fields will need more engineering effort to determine the type of each field (string, number, bool, struct) and output accordingly. For csv output it makes the columns and their sequence not predictable, and also less possible for further processing via spreadsheet formula. 
I think outputting all available metadata can be the next feature we provide. But for now I'm not sure what the use cases will be (besides troubleshooting during constraint development, which can be done with json format now), and not sure what's the best way to handle this.

cli/scorecard/cmd.go
```diff
@@ -29,17 +30,20 @@ func init() {
 
 	Cmd.Flags().StringVar(&flags.outputPath, "output-path", "", "Path to directory to contain scorecard outputs. Output to console if not specified")
 
-	Cmd.Flags().StringVar(&flags.outputFormat, "output-format", "", "Format of scorecard outputs, can be txt, json or csv, default is txt")
+	Cmd.Flags().StringVar(&flags.outputFormat, "output-format", "txt", "Format of scorecard outputs, can be txt, json or csv")
 	viper.SetDefault("output-format", "txt")
 	viper.BindPFlag("output-format", Cmd.Flags().Lookup("output-format"))
 
+	Cmd.Flags().StringSliceVar(&flags.metadataFields, "output-metadata-fields", []string{}, "List of comma delimited violation metadata fields of string type to include in output. Works when --output-format is txt or csv. By default no metadata fields in output when --output-format is txt or csv. All metadata will be in output when --output-format is json.")
```
morgante: For consistency, if output-metadata-fields is set I think we should *only* output the selected fields if the output format is `json`. This way `output-metadata-fields` will have consistent behavior for all formats and the only difference is the default (whether metadata is output at all).

PR #555, opened 2020-02-03
Remove 'latest' default from initialClusterVersion

mamaso:
initialClusterVersion 'latest' causes errors when used with the releaseChannel field.
I think it would be better to remove the default, allowing the GKE API to choose what version to provide if unspecified by the cluster template.


ocsig:
Thank you for your contribution!

PR #560, opened 2020-02-06
Added serviceAccountEmailAddress to output fields

davidebelloni:
For example needed to create an instance and give to CloudSQL' serviceAccount the permission to write on a bucket


ocsig:
@davidebelloni Thank you for your contribution. FYI: Only repository admins and added collaborators can trigger CloudBuild tests for security reasons.

davidebelloni:
Hi,
thanks ... I've tried ;)
I cannot see also the error on the failed check

ocsig:
There is an unrelated issue in one of the tests for CloudSQL, I've seen it earlier already. Tracking that: #562

PR #564, opened 2020-02-07
Added billing iam test account ID to Cloud Build config

kunalkg11:
1. Added the env variable for billing iam test account ID
2. Added the billing iam test account ID as a local variable


morgante:
@kunalkg11 ```Error 400: update not yet supported for GitHub App Triggers```

Please file a bug with Terraform team.

PR #573, opened 2020-02-13
DB and user name collision

davidebelloni:
Permit multiple CloudSQL in the same deployment with same database name or same users

### DIFF
--- a/dm/CHANGELOG.md
+++ b/dm/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### 18.02.2020
 
-- Update cloudsql template to support multiple instances in a deployment with the same names for databases and users [#573](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/573)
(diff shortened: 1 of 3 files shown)


rick-pri:
This will delete existing resources with  the previous naming scheme.  Probably not a wise move with databases.

davidebelloni:
Hi,
you are absolutely right but at the moment the "name convention" does not permit multiple databases with the same name in the same deployment

I think that an ordered number in the name solve the problem, but I'm afraid about possible reordering

rick-pri:
I've raised **[this](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/578)** issue/feature request.  If we could abandon the databases then rename them and then acquire them again with the new naming convention that would work better for everyone (an even better solution would be having a way to update the resource names, I guess).

ocsig:
@davidebelloni Thanks for your contribution. Please make sure you are changing the version in the Schema file. ( as @rick-pri pointed out this is a greaking change.) Please also add a note to the Changelog file. I will take a look at the tests.

@rick-pri You are right about the fack this is a breaking change. We are trying to singal this in the schema version, so are users adopt this template with care.

davidebelloni:
@ocsig done

ocsig:
Test fails on a known (unrelated) issue.

PR #575, opened 2020-02-14
Scorecard - get real asset ancestry path

katze120:
Get real asset ancestry path from CAI exports, so that:

- "target" in the constraints can be effective
- users can optionally output input.asset.ancestry_path in constraint templates as metadata

This PR obtains real input.asset.ancestry_path from CAI data and enables scorecard to output it in violation metadata if this is specified in constraint templates.


cli/scorecard/violations.go
```diff
@@ -126,35 +126,41 @@ func getViolations(inventory *InventoryConfig, config *ScoringConfig) (*validato
 
 // converts raw JSON into Asset proto
 func getAssetFromJSON(input []byte) (*validator.Asset, error) {
-	asset := tfconverter.Asset{}
+	var asset map[string]interface{}
```
morgante: I'd like to avoid moving from a defined type to a generic map. If the change that's needed is simply having `Ancestors` available, can we either pull the type from somewhere else (config validator itself?) or update the converters repo?

cli/scorecard/violations.go
```diff
@@ -126,35 +126,41 @@ func getViolations(inventory *InventoryConfig, config *ScoringConfig) (*validato
...
-	// fmt.Printf("Asset parent: %v\n", pbAsset.GetResource().GetParent())
-	return "organization/0/project/test", nil
+	ancestors := pbAsset.Ancestors
+	cnt := len(ancestors)
+	revAncestors := make([]string, len(ancestors))
+	for idx := 0; idx < cnt; idx++ {
+		revAncestors[cnt-idx-1] = ancestors[idx]
+	}
+	ancestorPath := strings.Join(revAncestors, "/")
+	if ancestorPath == ""{
```
morgante:
```suggestion
	if ancestorPath == "" {
```

cli/scorecard/violations.go
```diff
@@ -126,35 +126,41 @@ func getViolations(inventory *InventoryConfig, config *ScoringConfig) (*validato
...
 	}
-
-	Log.Debug("Asset converted", "name", asset.Name, "ancestry", pbAsset.GetAncestryPath())
-
+	Log.Debug("Asset converted", "name", asset["Name"], "ancestry", pbAsset.GetAncestryPath())
 	return pbAsset, nil
 }
 
 // looks up the ancestry path for a given asset
 func getAncestryPath(pbAsset *validator.Asset) (string, error) {
```
morgante: Could we add a link here? https://github.com/forseti-security/config-validator/blob/6cce12f5cd0e089a5371277384d9379d17b38288/pkg/asset/asset.go#L95

briantkennedy:
FCV has a few more changes to go then it should be stable for quite a while

cli/scorecard/violations.go
katze120: asset is a temporary variable for data type conversion purpose, immediately this is then converted to pbAsset which is defined proto message before being used further.

I'd like to remove dependency with tf validator, because the only thing we used there is the type. External dependencies add additional complexity in code development and testing. Considering we in near future will have similar updates for new data types and fields from new CAI-covered resources such as org policy, access levels, etc, updates to tf validator can be handled separately with different priority if such data will not be available in tfplan yet 

"ancestors" field gets lost during conversion to tfconverter.Asset, so it's not possible to subsequently get from validator.Asset. 
It's possible to get from the original json, via a similar process (converting to map[string]interface{} then get the "ancestors" field then join the reversed array), which overlaps with the approach in this PR and is more long winded. 

Forseti converts CAI json data to proto with directly, ignoring unknown fields (the only field in CAI but not in proto now is iam_policy.audit_configs), similar to the approach in this PR. ([code reference](https://github.com/forseti-security/forseti-security/blob/dev/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py#L94))
cft validate subcommand makes a copy of Asset struct definition within cft ([code reference](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/cli/validation/cai/cai.go)). I have not fully understood what validate subcommand is going, but if we definitely want to maintain a copy of Asset struct, we can refactor this to a common folder and have both scorecard and validate subcommands use this.

cli/scorecard/violations.go
katze120: this can be done after #577 is merged

morgante:
@katze120 Looks like this needs a rebase.

cli/scorecard/violations.go
katze120: done in 61a3f4e

cli/scorecard/violations.go
katze120: this line no longer exists

cli/scorecard/violations.go
katze120: tf validator has additional incompatibility with the new config validator when handling k8s resources. 
scorecard will return `failed to convert asset to admission request: resource.data field not found` with the current master code when handling k8s resources
This is because new config validator tries to convert k8s resource to admission controller format (seems cool for gatekeeper compatibility) and during the conversion it checks for resource.data which exists in CAI but the fields are reorganized in tf asset
Hence I recommend to stick to `map[string]interface{}` for as transient placeholder during intermediate conversion so as to be flexible with changes in CAI and config validator.

PR #588, opened 2020-02-26
Scorecard - add some tests 

katze120:
From 0% to 32.7% coverage.


cli/testdata/scorecard/cai-dir/resource_inventory.json
```diff
@@ -0,0 +1 @@
+{"name":"//storage.googleapis.com/test-bucket-public","asset_type":"storage.googleapis.com/Bucket","resource":{"version":"v1","discovery_document_uri":"https://www.googleapis.com/discovery/v1/apis/storage/v1/rest","discovery_name":"Bucket","parent":"//cloudresourcemanager.googleapis.com/projects/1234","data":{"acl":[],"billing":{},"cors":[],"defaultObjectAcl":[],"encryption":{},"etag":"CAU=","iamConfiguration":{"bucketPolicyOnly":{"enabled":true,"lockedTime":"2019-08-05T09:26:23.996Z"},"uniformBucketLevelAccess":{"enabled":true,"lockedTime":"2019-08-05T09:26:23.996Z"}},"id":"test-bucket-public","kind":"storage#bucket","labels":{},"lifecycle":{"rule":[]},"location":"US","locationType":"multi-region","logging":{},"metageneration":5,"name":"test-bucket-public","owner":{},"projectNumber":4567,"retentionPolicy":{},"selfLink":"https://www.googleapis.com/storage/v1/b/test-bucket-public","storageClass":"MULTI_REGIONAL","timeCreated":"2019-05-07T09:26:23.996Z","updated":"2019-05-07T15:11:21.219Z","versioning":{},"website":{}}},"ancestors":["projects/1234","folders/2345","organizations/56789"]}
```
morgante: nit. It'd be good to format test data for readability.

cli/scorecard/violations_test.go
```diff
@@ -0,0 +1,172 @@
...
+		})
+	}
+}
+
+var defaultGetAssetFromJSONs = map[string]string{
+	"testResourceJSON":  testResourceJSON,
+	"testIamPolicyJSON": testIamPolicyJSON,
+}
+
+var testResourceJSON = `
```
morgante: We should pull these from `testdata/`.

cli/testdata/scorecard/policy-library/lib/constraints.rego
```diff
@@ -0,0 +1,36 @@
+#
```
morgante: We shouldn't need the `lib/` directory for testdata.

katze120: That's required to initialize the validator because `lib` is hardcoded here: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/cli/scorecard/score.go#L50
.. which makes sense because we also want users to have it.

I agree that the files inside `lib` are not needed as of now because the test template does not use any lib functions. I'm going to put a .gitignore file there so that this empty directory can be committed

cli/scorecard/proto_test.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			t.Fatal("unexpected error", err)
+		}
+		if gotStr != wantedStr {
+			t.Errorf("wanted %s, got %s", wantedStr, gotStr)
+		}
+	})
+}
+
+var testIamPolicyAuditLogsJSON = `
```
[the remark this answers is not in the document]
katze120: done!

cli/scorecard/violations_test.go
katze120: done!

cli/testdata/scorecard/cai-dir/resource_inventory.json
katze120: This is a string comparison test and the wanted output is a single-line json

morgante: Got it. For the future, we should probably use json diffs.

PR #598, opened 2020-03-05
Add Config Connector Solutions + project-hierarchy

BrianPi:
Adding directory structure and READMEs for Config Connector Solutions, as well as first solution, project-hierarchy.

PR #599, opened 2020-03-05
Simple project (KCC Solution)

trodge:
This is a proposed implementation of a sample kpt package for creating a project in GCP using Config Connector CRDs

### DIFF
--- a/config-connector/solutions/projects/kpt/simple-project/Kptfile
+++ b/config-connector/solutions/projects/kpt/simple-project/Kptfile
@@ -1,0 +1,36 @@
+apiVersion: kpt.dev/v1alpha1
+kind: Kptfile
+metadata:
+  name: simple-project
+packageMetadata:
+  shortDescription: A simple project
+openAPI:
+  definitions:
+    io.k8s.cli.setters.org-id:
+      description: organization ID for associated services
+      x-k8s-cli:
+        setter:
+          name: org-id
+          value: "${ORG_ID?}"
+          setBy: PLACEHOLDER
+    io.k8s.cli.setters.project-id:
+      description: ID of project
+      x-k8s-cli:
+        setter:
+          name: project-id
+          value: simple-project
+          setBy: package-default
+    io.k8s.cli.setters.project-name:
+      description: display name of project
+      x-k8s-cli:
+        setter:
+          name: project-name
+          value: simple-project
+          setBy: package-default
+    io.k8s.cli.setters.billing-account:
+      description: ID of billing account
+      x-k8s-cli:
+        setter:
+          name: billing-account
+          value: "${BILLING_ACCOUNT_ID?}"
+          setBy: PLACEHOLDER
--- a/config-connector/solutions/projects/kpt/simple-project/Kptfile
+++ b/config-connector/solutions/projects/kpt/simple-project/Kptfile
@@ -20,13 +20,6 @@ openAPI:
           name: project-id
           value: simple-project
           setBy: package-default
-    io.k8s.cli.setters.project-name:
-      description: display name of project
-      x-k8s-cli:
-        setter:
-          name: project-name
-          value: simple-project
-          setBy: package-default
     io.k8s.cli.setters.billing-account:
       description: ID of billing account
       x-k8s-cli:
--- a/config-connector/solutions/projects/kpt/simple-project/README.md
+++ b/config-connector/solutions/projects/kpt/simple-project/README.md
@@ -15,9 +15,16 @@ Simple Project
   with your billing account and organization id numbers. You can do this with kpt setters or manually. To use kpt setters:
   
   ```
-  kpt cfg set . billing-account <insert billing account here>
-  kpt cfg set . org-id <insert organization id here>
+  kpt cfg set . billing-account VALUE 
+  kpt cfg set . org-id VALUE 
   ```
+
+  To set the project ID:
+  ```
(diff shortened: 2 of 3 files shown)


trodge:
@googlebot I consent.

BrianPi:
@googlebot I consent.

config-connector/solutions/projects/kpt/simple-project/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+  simple-project
+
+# SYNOPSIS
+
+  Config Connector compatible YAML files to create
+  a project in an organization. Make sure your Config Connector
+  service account has the project creator role for your organization.
+
+  In order to use, replace the `${BILLING_ACCOUNT_ID?}` and `${ORG_ID?}` values
```
maqiuyujoyce: You may want to add commands to set billing account ID and org ID here to better describe the usage.

config-connector/solutions/projects/kpt/simple-project/project.yaml
```diff
@@ -0,0 +1,22 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
```
maqiuyujoyce: Could you add the copyright above?

maqiuyujoyce:
@morgante They are two separate solutions under the same solution area, projects. #598 is for project hierarchy and this one is for simple project.

AlexBulankou:
> @morgante They are two separate solutions under the same solution area, projects. #598 is for project hierarchy and this one is for simple project.

@trodge, @maqiuyujoyce , @BrianPi  If you agree, I'd say let's have project-hierarchy go first since it is further along and then this one will have to merge before submitting?

config-connector/solutions/projects/kpt/simple-project/README.md
```diff
@@ -0,0 +1,31 @@
...
+  with your billing account and organization id numbers. You can do this with kpt setters or manually. To use kpt setters:
+  
+  ```
+  kpt cfg set billing-account <insert billing account here>
```
maqiuyujoyce: The dir param is missing in the commands.

trodge: Yes, you are correct, I will edit that.

config-connector/solutions/projects/kpt/simple-project/project.yaml
trodge: Yes, added.

config-connector/solutions/projects/kpt/simple-project/README.md
trodge: Added, now fixing syntax in next issue.

config-connector/solutions/projects/kpt/simple-project/Kptfile
```diff
@@ -0,0 +1,36 @@
...
+          value: "${ORG_ID?}"
+          setBy: PLACEHOLDER
+    io.k8s.cli.setters.project-id:
+      description: ID of project
+      x-k8s-cli:
+        setter:
+          name: project-id
+          value: simple-project
+          setBy: package-default
+    io.k8s.cli.setters.project-name:
```
morgante: Can we default this to be the same as project-id?

trodge: Yes, currently they both have value of `simple-project`. Should I make them use the same setter and therefore always have the same value?

morgante: The ideal would be that setting `project-id` will set both, while setting `project-name` would override the project name only.

trodge: The best way I can think of to implement this behavior with kpt would be to have them both be set by the project-id setter and make a comment in project.yaml noting that the spec.name can be changed independently of the metadata.name.

morgante: Could we ask kpt team about multi-setters or something like this? It seems like a reasonable use case.

trodge: I have created an issue for it [here](https://github.com/GoogleContainerTools/kpt/issues/430).

trodge: @morgante in the mean time I have made them both use the same setter. Please take a look when you get the chance.

config-connector/solutions/projects/kpt/simple-project/README.md
```diff
@@ -0,0 +1,49 @@
...
+# REQUIREMENTS
+  A working cluster with Config Connector installed.
+  
+  Cloud Resource Manager API and Cloud Billing API enabled in the project where Config Connector is installed.
+  
+  The cnrm-system service account must have `roles/resourcemanager.projectCreator` in your organization and `roles/billing.user` for your billing account.
+  
+# USAGE
+  In order to use, replace the `${BILLING_ACCOUNT_ID?}` and `${ORG_ID?}` values with your billing account and organization id. You can do this with kpt setters or by manually editing the YAML files. To use kpt setters:
+  ```
```
maqiuyujoyce: "You can do this with kpt setters or by manually editing the YAML files. To use kpt setters:" -> "You can do this with kpt setters:"

Let's not suggest manual edits if it is not required.

trodge:
@AlexBulankou @maqiuyujoyce I believe I have addressed all comments, take a look when you get a chance.

PR #559, opened 2020-02-04
Removing Schema self reference imports

ocsig:

PR #604, opened 2020-03-12
Revert changes from #559 and import the subnetwork.py template

frots:
During fixing issue #559 this template was incorrectly changed braking existing templates. This PR reverts the change.


ocsig:
Good catch! Thanks for the contribution!

PR #605, opened 2020-03-12
Add Config Connector solution member-iam

maqiuyujoyce:
Adding a new solution for Config Connector. This solution comes from the TF CFT member_iam [example](https://github.com/terraform-google-modules/terraform-google-iam/tree/master/examples/member_iam)/[submodule](https://github.com/terraform-google-modules/terraform-google-iam/tree/master/modules/member_iam).

### DIFF
--- a/config-connector/solutions/iam/kpt/member-iam/0-namespace.yaml
+++ b/config-connector/solutions/iam/kpt/member-iam/0-namespace.yaml
@@ -12,17 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: iam.cnrm.cloud.google.com/v1beta1
-kind: IAMPolicyMember
+apiVersion: v1
+kind: Namespace
 metadata:
-  name: project-iam-member
-  namespace: member-iam-solution
-spec:
-  # Replace the ${PROJECT_ID?} below with your desired project ID.
-  member: serviceAccount:member-iam-test@${PROJECT_ID?}.iam.gserviceaccount.com # {"$ref":"#/definitions/io.k8s.cli.substitutions.project-id-in-sa-sub"}
-  role: roles/compute.networkAdmin # {"$ref":"#/definitions/io.k8s.cli.substitutions.role-sub"}
-  resourceRef:
+  annotations:
     # Replace the ${PROJECT_ID?} below with your desired project ID.
-    apiVersion: resourcemanager.cnrm.cloud.google.com/v1beta1
-    kind: Project
-    external: projects/${PROJECT_ID?} # {"$ref":"#/definitions/io.k8s.cli.substitutions.project-id-sub"}
+    cnrm.cloud.google.com/project-id: ${PROJECT_ID?} # {"$ref":"#/definitions/io.k8s.cli.setters.project-id"}
+  name: member-iam-solution
--- a/config-connector/solutions/iam/kpt/member-iam/Kptfile
+++ b/config-connector/solutions/iam/kpt/member-iam/Kptfile
@@ -1,0 +1,63 @@
+apiVersion: kpt.dev/v1alpha1
+kind: Kptfile
+metadata:
+  name: .
+packageMetadata:
+  shortDescription: sample description
+openAPI:
+  definitions:
+    io.k8s.cli.setters.project-id:
+      x-k8s-cli:
+        setter:
+          name: project-id
+          value: ${PROJECT_ID?}
+          setBy: PLACEHOLDER
+      description: The target project where you want to create service accounts and
+        grant permissions.
+    io.k8s.cli.substitutions.project-id-in-sa-sub:
+      x-k8s-cli:
+        substitution:
+          name: project-id-in-sa-sub
+          pattern: serviceAccount:SERVICE_ACCOUNT_NAME_SETTER@PROJECT_ID_SETTER.iam.gserviceaccount.com
+          values:
+          - marker: PROJECT_ID_SETTER
+            ref: '#/definitions/io.k8s.cli.setters.project-id'
+          - marker: SERVICE_ACCOUNT_NAME_SETTER
+            ref: '#/definitions/io.k8s.cli.setters.service-account-name'
+    io.k8s.cli.substitutions.project-id-sub:
+      x-k8s-cli:
+        substitution:
+          name: project-id-sub
+          pattern: projects/PROJECT_ID_SETTER
+          values:
+          - marker: PROJECT_ID_SETTER
+            ref: '#/definitions/io.k8s.cli.setters.project-id'
+    io.k8s.cli.setters.policy-member-name:
+      x-k8s-cli:
+        setter:
+          name: policy-member-name
+          value: project-iam-member
+          setBy: package-default
+      description: The name of the instantiated KCC IAMPolicyMember resource.
+    io.k8s.cli.setters.service-account-name:
+      x-k8s-cli:
+        setter:
+          name: service-account-name
+          value: member-iam-test
+          setBy: package-default
+      description: The name of the new IAM service account.
+    io.k8s.cli.setters.role:
+      x-k8s-cli:
+        setter:
+          name: role
+          value: compute.networkAdmin
+          setBy: package-default
+      description: The role you want to grant the service account with.
+    io.k8s.cli.substitutions.role-sub:
+      x-k8s-cli:
+        substitution:
+          name: role-sub
+          pattern: roles/ROLE_SETTER
+          values:
+          - marker: ROLE_SETTER
+            ref: '#/definitions/io.k8s.cli.setters.role'
--- a/config-connector/solutions/iam/kpt/member-iam/Kptfile
+++ b/config-connector/solutions/iam/kpt/member-iam/Kptfile
@@ -3,7 +3,7 @@ kind: Kptfile
 metadata:
   name: .
 packageMetadata:
-  shortDescription: sample description
+  shortDescription: Create a service account and grant it permissions to access the desired project.
 openAPI:
   definitions:
     io.k8s.cli.setters.project-id:
--- a/config-connector/solutions/iam/kpt/member-iam/Kptfile
+++ b/config-connector/solutions/iam/kpt/member-iam/Kptfile
@@ -43,14 +43,6 @@ openAPI:
       x-k8s-cli:
         setter:
           name: role
-          value: compute.networkAdmin
+          value: roles/compute.networkAdmin
           setBy: package-default
       description: The role you want to grant the service account with.
-    io.k8s.cli.substitutions.role-sub:
-      x-k8s-cli:
-        substitution:
-          name: role-sub
-          pattern: roles/ROLE_SETTER
-          values:
-          - marker: ROLE_SETTER
-            ref: '#/definitions/io.k8s.cli.setters.role'
--- a/config-connector/solutions/iam/kpt/member-iam/Kptfile
+++ b/config-connector/solutions/iam/kpt/member-iam/Kptfile
@@ -32,13 +32,6 @@ openAPI:
           values:
           - marker: PROJECT_ID_SETTER
             ref: '#/definitions/io.k8s.cli.setters.project-id'
-    io.k8s.cli.setters.policy-member-name:
-      x-k8s-cli:
-        setter:
-          name: policy-member-name
-          value: project-iam-member
-          setBy: package-default
-      description: The name of the instantiated KCC IAMPolicyMember resource.
     io.k8s.cli.setters.service-account-name:
       x-k8s-cli:
         setter:
--- a/config-connector/solutions/iam/kpt/member-iam/README.md
+++ b/config-connector/solutions/iam/kpt/member-iam/README.md
@@ -1,0 +1,84 @@
+Member IAM
+==================================================
+
+# NAME
+
+  member-iam
+
+# SYNOPSIS
+
+  Config Connector compatible YAML files to create a service account in your desired project, and grant it a specific role (defaults to `compute.networkAdmin`) in the project.
+
+  You can also create any number of service accounts and grant them any number of roles by altering and applying the altered config YAMLs.
+
+# CONSUMPTION
+
+  Using [kpt](https://googlecontainertools.github.io/kpt/):
+
+  ```
+  kpt pkg get https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit.git/config-connector/solutions/iam/kpt/member-iam member-iam
+  ```
+
+# REQUIREMENTS
+
+  A working Config Connector cluster using a service account 
+  (`cnrm-system@[PROJECT_ID].iam.gserviceaccount.com`) with the following 
+  roles in your desired project (it doesn't need to be the project where you 
+  installed Config Connector):
+
+  - roles/resourcemanager.projectIamAdmin
+  - roles/iam.serviceAccountAdmin
+
+# USAGE
+
+  Replace `${PROJECT_ID?}` with your desired project ID value from 
+  within this directory:
+
+  ```
+  kpt cfg set . project-id VALUE
+  ```
+
+  Once the project ID is set in the configs, simply apply the YAMLs:
+
+  ```
+  kubectl apply -f .
+  ```
+
+## OPTIONAL WORKFLOW
+
+  Optionally, you may want to create a different service account, grant a 
+  different role to the service account, or grant multiple roles to the 
+  service account, etc. Here are the instructions.
+
+**Create a Different Service Account**
+
+  Change the service account name and apply the YAMLs:
+
+  ```
+  kpt cfg set . service-account-name VALUE
+  kubectl apply -f .
+  ```
+
+**Grant a Different Role**
+  
+  Change the role([predefined GCP IAM roles](https://cloud.google.com/iam/docs/understanding-roles#predefined_roles)) and apply the YAMLs:
+
+  ```
+  kpt cfg set . role VALUE
+  kubectl apply -f .
+  ```
+
+**Grant Multiple Roles**
+
+  Change the KCC resource name and the role each time before you grant a new
+  role to the service account:
+
+  ```
+  kpt cfg set . policy-member-name VALUE_1
+  kpt cfg set . role VALUE_2
+  kubectl apply -f iampolicymember.yaml
+  ```
+
+# LICENSE
+
+  Apache 2.0 - See [LICENSE](/LICENSE) for more information.
--- a/config-connector/solutions/iam/kpt/member-iam/README.md
+++ b/config-connector/solutions/iam/kpt/member-iam/README.md
@@ -9,8 +9,6 @@ Member IAM
 
   Config Connector compatible YAML files to create a service account in your desired project, and grant it a specific role (defaults to `compute.networkAdmin`) in the project.
 
-  You can also create any number of service accounts and grant them any number of roles by altering and applying the altered config YAMLs.
-
 # CONSUMPTION
 
   Using [kpt](https://googlecontainertools.github.io/kpt/):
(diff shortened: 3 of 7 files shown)


config-connector/solutions/iam/kpt/member-iam/Kptfile
```diff
@@ -0,0 +1,63 @@
...
+  definitions:
+    io.k8s.cli.setters.project-id:
+      x-k8s-cli:
+        setter:
+          name: project-id
+          value: ${PROJECT_ID?}
+          setBy: PLACEHOLDER
+      description: The target project where you want to create service accounts and
+        grant permissions.
+    io.k8s.cli.substitutions.project-id-in-sa-sub:
```
AlexBulankou: Why do we need this separate setter? Can we not set project_id and service_account_name from project-id and service-account-name setters?

maqiuyujoyce: "project-id-in-sa-sub" is not a setter, but a string substitution. With this subst, we can substitute project id in the service account. You can find more information here: https://googlecontainertools.github.io/kpt/reference/cfg/create-subst/

maqiuyujoyce:
There is already a README at config-connector/solutions level in #598, I think that's good enough. But yes we should add some descriptions at solution area level.

config-connector/solutions/iam/kpt/member-iam/README.md
```diff
@@ -0,0 +1,84 @@
+Member IAM
+==================================================
+
+# NAME
+
+  member-iam
+
+# SYNOPSIS
+
+  Config Connector compatible YAML files to create a service account in your desired project, and grant it a specific role (defaults to `compute.networkAdmin`) in the project.
```
BrianPi: Is it important to call out what the default role is here? I'd assume that the package consumer can already see it in YAML/setter form, and doesn't need to be told.

config-connector/solutions/iam/kpt/member-iam/README.md
```diff
@@ -0,0 +1,84 @@
...
+  Using [kpt](https://googlecontainertools.github.io/kpt/):
+
+  ```
+  kpt pkg get https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit.git/config-connector/solutions/iam/kpt/member-iam member-iam
+  ```
+
+# REQUIREMENTS
+
+  A working Config Connector cluster using a service account 
+  (`cnrm-system@[PROJECT_ID].iam.gserviceaccount.com`) with the following
```
BrianPi: A working Config Connector cluster depends on using the cnrm-system@[PROJECT_ID].iam.gserviceaccount.com service account... since we're not applying the role to this account, I don't think it's important to mention.

config-connector/solutions/iam/kpt/member-iam/README.md
```diff
@@ -0,0 +1,84 @@
...
+  kpt cfg set . project-id VALUE
+  ```
+
+  Once the project ID is set in the configs, simply apply the YAMLs:
+
+  ```
+  kubectl apply -f .
+  ```
+
+## OPTIONAL WORKFLOW
```
BrianPi: I feel like we may want to hold off on an optional workflow section, or make it more targeted instead of just appending it to # USAGE... I feel that as is it doesn't help too much, and the **Grant Multiple Roles** section especially seems unclear to me.

config-connector/solutions/iam/kpt/member-iam/README.md
```diff
@@ -0,0 +1,84 @@
...
+# USAGE
+
+  Replace `${PROJECT_ID?}` with your desired project ID value from 
+  within this directory:
+
+  ```
+  kpt cfg set . project-id VALUE
+  ```
+
+  Once the project ID is set in the configs, simply apply the YAMLs:
```
BrianPi: I feel like we should rephrase this to make it clear that it is the customer's option to do this when they feel ready. They may want to apply additional configuration.

config-connector/solutions/iam/kpt/member-iam/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+# NAME
+
+  member-iam
+
+# SYNOPSIS
+
+  Config Connector compatible YAML files to create a service account in your desired project, and grant it a specific role (defaults to `compute.networkAdmin`) in the project.
+
+  You can also create any number of service accounts and grant them any number of roles by altering and applying the altered config YAMLs.
```
BrianPi: I'm not sure we should explicitly tell the customer to do this. It makes it easy to lose track of what they've applied; the mapping between kpt package and deployment is lost.

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: In order to create a new service account in the desired project and apply the roles specified in the yaml file to the new service account, the cnrm-system@ gsa MUST have the following permissions. So that's a requirement. We don't assume the desired the project is the same as the project where the customer has installed KCC.

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: Yes and no. After reading your comment, I rethought about it. I think calling out the default role is useful but just calling out the default role is not sufficient.

Why useful - Customers will probably read READMEs at the first trial, but they will unlikely read YAMLs. If compute.networkAdmin is a sensitive role to them, I'd prefer them know it sooner. 

Why not sufficient - I think what's missing in the doc is the output of the get commands (like list-setters), because customers won't know what is the right output if they've never used this package/kpt tool before. Solutions are for customers to get onboard, so we should expect customers are not sophisticated enough to figure out information they want by themselves. Providing the expected output for customers to compare with will ensure they are on the right track. It's just like how TF CFT READMEs do ([example](https://github.com/terraform-google-modules/terraform-google-iam/tree/master/modules/folders_iam)).

I'm in general a fan of providing redundant information to customers to avoid them missing important info. Even if it means harder maintenance. We all like TF documentation, and one reason I find it's useful is that they provide sufficient information even if sometimes it is redundant. E.g. for the default field in the resources, it is defined in the schema (go code), specified in the doc (maintained separately), and is sometimes also mentioned in the samples.

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: I think for getting onboard, we want to be more prescriptive. I treat it like quickstart guide. What do you think?

config-connector/solutions/iam/kpt/member-iam/README.md
BrianPi: That's why I think that explicitly calling out the service account name is confusing; which project is [PROJECT_ID] referring to?

config-connector/solutions/iam/kpt/member-iam/README.md
BrianPi: I think it's appropriate to push the customer towards best practices wherever possible, but if we're too pushy, we also risk pushing the customer away from best practices. This is especially true in this situation, where we're engaged in a new initiative and the package manager we're using isn't stable yet.

config-connector/solutions/iam/kpt/member-iam/Kptfile
AlexBulankou: Got it, thanks.

config-connector/solutions/iam/kpt/member-iam/Kptfile
```diff
@@ -0,0 +1,63 @@
...
+            ref: '#/definitions/io.k8s.cli.setters.service-account-name'
+    io.k8s.cli.substitutions.project-id-sub:
+      x-k8s-cli:
+        substitution:
+          name: project-id-sub
+          pattern: projects/PROJECT_ID_SETTER
+          values:
+          - marker: PROJECT_ID_SETTER
+            ref: '#/definitions/io.k8s.cli.setters.project-id'
+    io.k8s.cli.setters.policy-member-name:
```
trodge: Is there any value in making this a setter? It seems that in the optional workflow the user will need to edit yaml files manually anyway in order to use this since resource names need to be unique within the project.

config-connector/solutions/iam/kpt/member-iam/README.md
trodge: I think we can be prescriptive in areas where we are certain of best practice and less prescriptive where we are not.

config-connector/solutions/iam/kpt/member-iam/Kptfile
maqiuyujoyce: This field is made to be a setter because we want customers to use it to change the resource name. They don't need to manually update in the yaml file. I think I explicitly described it in the optional workflow. Did you find it confusing?

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: As we discussed offline, let's just leave these kind details at the comfort level of the solution dev and update them later if we receive any customer feedback.

In regard to whether we are going to list all the setters in the README or not, we'll decide later.

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: As we discussed offline, let's just leave this kind details at the comfort level of the solution dev and update them later if we receive any customer feedback.

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: As we discussed offline, let's just leave this kind details at the comfort level of the solution dev and update them later if we receive any customer feedback.

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: As we discussed offline, let's call it Config Connector "cnrm-system" service account.

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: As we discussed offline, let's just make it another top level section in README.

maqiuyujoyce:
Hi Morgante, thanks a lot for your feedback. Here are my responses to your concerns.

> I think we need to rethink this solution as it seems quite complicated/confusing, especially relative to the underlying resources. Right now, it seems like a very thin wrapper around the resource which doesn't really provide much value.

Yes, it is a thin wrapper. As we discussed last Friday, we currently want to focus on providing simple/basic solutions and improve based on user feedback. So the main value for now is to be a Config Connector sample using kpt and a channel for user feedback.

> In particular, I'm very much not a fan of adding additional roles by renaming the kubernetes resource (essentially abandoning the old resource so it doesn't exist anywhere on disk).

You are right, it is not the best practice, but it is a working flow users can follow with kpt right now. Just want to clarify, users can still delete the resources using `kubectl delete [Kind] [Name]` directly even if the files are gone/overwritten locally on their machine. Our target users are sophisticated GCP users with decent amount of Kubernetes knowledge, so we assume they are able to manage the resources properly even if the solution is not perfect. An alternative way to do it is to copy the yaml and make manual changes, but it won't work with kpt setters then because it will mess up the values in the yaml files. Considering users are not kpt experts, the former workflow is more reasonable.

> At the very least, I think we should make it possible to add additional roles to a service account declaratively without losing the existing one.

I agree that the ideal scenario is that users can dynamically parametrize an arbitrary number of the "template" yamls and we do want to support it. But `kpt` doesn't support templating the yaml file and creating an arbitrary number of the parametrized copies of it now. That's why we suggested the manual overwrite/editing in our solution README. Did you consider manual workaround not preferred?

Other alternatives are using Helm (flexible, out of the scope of kpt), or using kpt fn (powerful, too heavy weight for platform admin to develop on top of) at this moment. I'll look into the other two options and see how we want to prioritize them. In general, they all have their own disadvantages, and the manual option may be the best way to go right now considering usability and flexibility.

AlexBulankou:
I agree that we should not be recommending to add additional roles by renaming the kubernetes resource and "(essentially abandoning the old resource so it doesn't exist anywhere on disk).". Rather than recommending the user to change the existing role, we can ask the user to copy and customize provided template for the new role. This will preserve the existing role.

morgante:
> Other alternatives are using Helm (flexible, out of the scope of kpt), or using kpt fn (powerful, too heavy weight for platform admin to develop on top of) at this moment.

I think we should strongly consider leveraging `kpt fn` directly. For example, there is a kpt fn to run Helm: https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/helm-template

I believe there's also a Jinja2 fn somewhere (or could easily be written).

maqiuyujoyce:
Could you clarify a bit, Morgante? Are you suggesting that we create a kpt fn that can understand Jinja2 and generate yamls out of it?
I think `kpt fn` is considered a break-glass for the functions we need. So in this example, it could be a kpt fn taking a list of <resource_name, member, role> values, parametrize the template yaml, and generate new yaml files.
We don't really want to create and maintain a generic kpt fn that dynamically parametrize any templates because that should be the feature in kpt/helm/etc, and will be out of scope of our solutions effort.

AlexBulankou:
@morgante ,  we will consider how to use `kpt fn` as expansion tool in future solutions. We don't want to use it for this solution - we only want to limit it to the single template and explain how customers can expand it manually and not introduce `kpt fn` for expansion. Can we agree to proceed without automatic expansion for this solution? i.e. without introducing `kpt fn` - as this is outside of what is in our toolbox for this solution.

morgante:
If we want to prioritize releasing something, we can do so. I just don't think the current solution has much, if any, practical value.

Regardless, I feel strongly that we should *not* encourage people to edit the resources on disk and lose declarative modeling. If we don't have a viable solution for managing multiple members, let's omit it entirely.

trodge:
> If we want to prioritize releasing something, we can do so. I just don't think the current solution has much, if any, practical value.
> 
> Regardless, I feel strongly that we should _not_ encourage people to edit the resources on disk and lose declarative modeling. If we don't have a viable solution for managing multiple members, let's omit it entirely.

One possible solution that would preserve declarative modeling would be to instruct the user to create  new IAMPolicyMember resource. To do this they would need to copy the contents of iampolicymember.yaml and either edit manually or create setters for the resource name, role, and member of each role/member pair they want to assign. Currently an IAMPolicyMember can only grant a single role to a single member.

If we want to be able to handle more complex scenarios with a single resource yaml we would need to use an IAMPolicy resource. This resource has two main limitations that would make it inappropriate for this solution:
 - It uses a list of bindings (each of which has a list of members and a role) which kpt does not currently (and possibly will never) support manipulating with setters.
- It will overwrite existing IAM policy configuration which users may have created by other means.

We are interested in pushing this solution forward using the IAMPolicyMember resource as a demonstration of what kpt setters are currently capable of, which we hope can provide some of the function [equivalent terraform module](https://github.com/terraform-google-modules/terraform-google-iam/tree/master/modules/member_iam). If customer feedback suggests that the functionality with setters is not sufficient to their needs, that will guide us toward either kpt fn or helm.

config-connector/solutions/iam/kpt/member-iam/Kptfile
trodge: To me, a setter signals that this is something we expect the customer to want to change. I was unclear as to why the user would be changing it. It appears that this was part of the optional workflow. I'm not sure if kpt setters are going to be useful at all in the case where a customer wants to apply an arbitrary set of roles to an arbitrary set of members.

morgante:
I agree that `IAMPolicy` is dangerous and probably shouldn't be pursued.

> We are interested in pushing this solution forward using the IAMPolicyMember resource as a demonstration of what kpt setters are currently capable of, which we hope can provide some of the function equivalent terraform module.

One of the key things about the module is the ability to accept multiple roles. If all the module did was take a given member and a given role, we probably wouldn't have released it.

> One possible solution that would preserve declarative modeling would be to instruct the user to create new IAMPolicyMember resource. To do this they would need to copy the contents of iampolicymember.yaml and either edit manually or create setters for the resource name, role, and member of each role/member pair they want to assign. 

Right, this is still rather cumbersome to the point that I'd suggest we omit it entirely. Since we don't have any good current solution, users can still `kpt get` the package multiple times for different roles.

maqiuyujoyce:
I removed the optional workflow in the README and we no longer suggest customers make manual edits. We will look into how to enable the the automatic expansion in the future, but for now let's just limit it to a single yaml.

I also used a new namespace to map to the desired project that customer specifies. This is Config Connector's best practices for managing resources in different projects.

config-connector/solutions/iam/kpt/member-iam/README.md
```diff
@@ -0,0 +1,62 @@
...
+# USAGE
+
+  Replace `${PROJECT_ID?}` with your desired project ID value from 
+  within this directory:
+
+  ```
+  kpt cfg set . project-id VALUE
+  ```
+
+  _Optionally_, you can also change the service acocunt name and role
```
AlexBulankou, line 38: account (typo)

AlexBulankou:
Following up on the earlier point from @morgante  - I also agree that we should not recommend change the resources on disk and thus orphan existing GCP configs. I believe latest changes from @maqiuyujoyce  addressed this problem.

config-connector/solutions/iam/kpt/member-iam/iampolicymember.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+apiVersion: iam.cnrm.cloud.google.com/v1beta1
+kind: IAMPolicyMember
+metadata:
+  name: project-iam-member
+  namespace: member-iam-solution
+spec:
+  # Replace the ${PROJECT_ID?} below with your desired project ID.
+  member: serviceAccount:member-iam-test@${PROJECT_ID?}.iam.gserviceaccount.com # {"$ref":"#/definitions/io.k8s.cli.substitutions.project-id-in-sa-sub"}
+  role: roles/compute.networkAdmin # {"$ref":"#/definitions/io.k8s.cli.substitutions.role-sub"}
```
trodge, line 23: We are using the fully qualified `roles/compute.networkAdmin` and a plain setter rather than a substitution in the other solutions.

config-connector/solutions/iam/kpt/member-iam/0-namespace.yaml
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    # Replace the ${PROJECT_ID?} below with your desired project ID.
+    cnrm.cloud.google.com/project-id: ${PROJECT_ID?} # {"$ref":"#/definitions/io.k8s.cli.setters.project-id"}
```
trodge, line 20: Should this namespace be named after the project-id with no annotation?

config-connector/solutions/iam/kpt/member-iam/Kptfile
```diff
@@ -0,0 +1,56 @@
...
+          name: project-id-sub
+          pattern: projects/PROJECT_ID_SETTER
+          values:
+          - marker: PROJECT_ID_SETTER
+            ref: '#/definitions/io.k8s.cli.setters.project-id'
+    io.k8s.cli.setters.service-account-name:
+      x-k8s-cli:
+        setter:
+          name: service-account-name
+          value: member-iam-test
```
trodge, line 39: Should perhaps have a different name, such as member-iam-service-account

maqiuyujoyce: I believe the resource kind is sufficient for customers to recognize that this is a service account, and we don't need to emphasize it in the name.

config-connector/solutions/iam/kpt/member-iam/iampolicymember.yaml
maqiuyujoyce: Done.

config-connector/solutions/iam/kpt/member-iam/README.md
maqiuyujoyce: Done.

config-connector/solutions/iam/kpt/member-iam/Kptfile
```diff
@@ -0,0 +1,48 @@
+apiVersion: kpt.dev/v1alpha1
+kind: Kptfile
+metadata:
+  name: .
+packageMetadata:
+  shortDescription: sample description
```
trodge, line 6: Should be a description of the package

maqiuyujoyce: Done.

PR #616, opened 2020-03-27
Update unmanaged_instance_group.py

davidebelloni:
BugFix UIG when setting network as selfLink


davidebelloni:
Hi,
why is this simple bugfix stalled?

PR #620, opened 2020-03-28
made namespace name match project-id

trodge:


BrianPi:
AFAIK, we decided that namespaces should be specific qualified names not tied to any customer set information. Is this still current?

PR #630, opened 2020-04-03
added mysql-ha solution

trodge:

PR #635, opened 2020-04-07
Mysql private solution

trodge:


config-connector/solutions/sql/kpt/mysql-private/README.md
```diff
@@ -0,0 +1,44 @@
...
+| region        | us-central1            | package-default | region of SQL instance | 1     |
+| user-name     | mysql-private-user     | package-default | name of SQL user       | 1     |
+
+  Configure setters using kpt as follows:
+  ```
+  kpt cfg set . NAME VALUE
+  ```
+  Setting placeholder values is required, changing package-defaults is optional.
+
+  For this package to work properly, the following resources must be in a ready state before the SQLInstance YAML is applied:
```
AlexBulankou, line 30: Is this the case that we need to ensure this ordering? Part of the promise of Config Connector is idempotency, so if this is the case, this would be inconsistent with the optimal experience that we trying to enable and we would need to explain the bug. Can you please verify what happens if apply all yamls in the same time?

config-connector/solutions/sql/kpt/mysql-private/README.md
```diff
@@ -0,0 +1,44 @@
...
+## SETTERS
+|     NAME      |         VALUE          |     SET BY      |      DESCRIPTION       | COUNT |
+|---------------|------------------------|-----------------|------------------------|-------|
+| database-name | mysql-private-database | package-default | name of SQL database   | 1     |
+| instance-name | mysql-private-instance | package-default | name of SQL instance   | 3     |
+| password      | ${PASSWORD?}           | PLACEHOLDER     | password of SQL user   | 1     |
+| region        | us-central1            | package-default | region of SQL instance | 1     |
+| user-name     | mysql-private-user     | package-default | name of SQL user       | 1     |
+
+  Configure setters using kpt as follows:
```
AlexBulankou, line 24: Can you please address the same feedback as in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/630 in this pull request? Thanks!

PR #641, opened 2020-04-08
Restrict building new dev tools images only when pushed to master

bharathkkb:
Build CFT dev tools image only when pushed to `master`


bharathkkb:
@morgante PTAL

PR #651, opened 2020-04-18
[KCC] Add the Go CLI to test Config Connector solutions

maqiuyujoyce:
Currently, it only runs the test for a single solution at a time. Improvements will be done in future PRs.

PR #656, opened 2020-04-23
[KCC] Remove the dependency on Go for test-cli users

maqiuyujoyce:
As we discussed in #651, I'm uploading the test-cli to remove the dependency on Go.


morgante:
In general we should avoid committing binary files. Could this be put in a GCS bucket instead?

maqiuyujoyce:
> In general we should avoid committing binary files. Could this be put in a GCS bucket instead?

Yeah, makes sense. Will do.

PR #666, opened 2020-05-01
[KCC] Add the testcase for simple-project solution

maqiuyujoyce:
* Updated the original solution to make project-id a required setter.
* Added the feature to append a random ID when the value needs to be globally unique.

### DIFF
--- a/config-connector/tests/ccs-test/cmd/root.go
+++ b/config-connector/tests/ccs-test/cmd/root.go
@@ -22,9 +22,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/ghodss/yaml"
+	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/config-connector/tests/ccs-test/util"
 	"github.com/spf13/cobra"
 )
 
@@ -39,6 +41,10 @@ const (
 
 var (
 	relativePath string
+
+	// Regex of the env vars that require a randomized suffix.
+	// It should be in the format of $ENV_VAR-$RANDOM_ID.
+	re = regexp.MustCompile(`\$(?P<EnvName>[A-Z]+|[A-Z]+[A-Z_]*[A-Z]+)(-\$RANDOM_ID)`)
 
 	rootCmd = &cobra.Command{
 		Use:   "ccs-test",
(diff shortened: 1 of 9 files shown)


maqiuyujoyce:
The updated test-cli will be uploaded after the PR is merged.

config-connector/tests/ccs-test/cmd/root.go
```diff
@@ -102,7 +108,14 @@ var (
 				return
 			}
 
-			realValues, err := finalizeValuesFromEnv(envValues, testValues)
+			// Generate the random IDs first.
+			randomId, err := util.GenerateRandomizedSuffix()
+			if err != nil {
+				log.Fatalf("error generating the randomized suffix for resource names: %v", err)
```
AlexBulankou, line 114: Should we return if error?

config-connector/tests/testcases/environments.template
```diff
@@ -11,5 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# RANDOM_ID should not be defined in the environments file.
+BILLING_ACCOUNT_ID: ${BILLING_ACCOUNT_ID?}
 DEFAULT_PROJECT_ID: ${DEFAULT_PROJECT_ID?}
 IAM_MEMBER: ${IAM_MEMBER?}
+ORG_ID: ${ORG_ID?}
+SIMPLE_PROJECT_ID: ${SIMPLE_PROJECT_ID?}
```
AlexBulankou, line 20: Now we have DEFAULT_PROJECT_ID and SIMPLE_PROJECT_ID. Why not just have a PROJECT_ID and use it for both? Worried that as we are adding more test cases a number of different SOMETHING_PROJECT_ID will grow.

config-connector/tests/ccs-test/cmd/root.go
maqiuyujoyce: Good point. Actually log.Fatalf() will print the error and exit with code 1 so the return above is never reached. Will clean them up later.

config-connector/tests/testcases/environments.template
maqiuyujoyce: SOMETHING_PROJECT_ID env vars (except for DEFAULT_PROJECT_ID, which is for an existing project) are used as the name prefix for new projects to create. Because project ID needs to be a globally unique value, they should be different for each test run for each solution. Currently there are only 4 solutions creating 5 projects in total so there will only be five SOMETHING_PROJECT_ID env vars. Will look into the improvements after finishing all the test cases.

PR #671, opened 2020-05-06
[KCC] Add the test case for owned-project solution

maqiuyujoyce:


maqiuyujoyce:
This test case still adds a new env var for the new project. Eventually users will only need to specify a NEW_PROJECT_ID_PREFIX, but this improvement will be done after all the test cases are backfilled.

config-connector/tests/testcases/environments.template
```diff
@@ -15,6 +15,8 @@
 # RANDOM_ID should not be defined in the environments file.
 BILLING_ACCOUNT_ID: ${BILLING_ACCOUNT_ID?}
 DEFAULT_PROJECT_ID: ${DEFAULT_PROJECT_ID?}
+FOLDER_ID: ${FOLDER_ID?}
 IAM_MEMBER: ${IAM_MEMBER?}
 ORG_ID: ${ORG_ID?}
+OWNED_PROJECT_ID: ${OWNED_PROJECT_ID?}
```
AlexBulankou, line 21: Same as my earlier comment - can we reuse PROJECT_ID for multiple tests - we have at this point 4 different values to indicate project ID, can we instead use PROJECT_ID across all of them?

maqiuyujoyce: Ack. I'll improve the PROJECT_ID related env vars later.

PR #677, opened 2020-05-08
Scorecard new feature - Support Org Policy, Access Policy asset types with new Config Validator version

katze120:


cli/scorecard/violations.go
```diff
@@ -74,7 +76,8 @@ func getDataFromFile(config *ScoringConfig, caiDirName string) ([]*validator.Ass
 	for _, objectName := range destinationObjectNames {
 		reader, err := os.Open(filepath.Join(caiDirName, objectName))
 		if err != nil {
-			return nil, err
+			Log.Debug("getDataFromFile", objectName, "File does not exist")
+			continue
```
katze120, line 80: Making each of the file non-mandatory with debug logs because the new asset type options org policy and access policy may not be widely known and used.

cli/scorecard/violations.go
```diff
@@ -126,21 +129,21 @@ func getViolations(inventory *InventoryConfig, config *ScoringConfig) (*validato
...
-	err = protoViaJSON(asset, pbAsset)
-	if err != nil {
-		return nil, errors.Wrapf(err, "converting asset %s to proto", asset["name"])
+	umar := &jsonpb.Unmarshaler{AllowUnknownFields: true}
+	if err := umar.Unmarshal(strings.NewReader(string(input)), pbAsset); err != nil {
+		err2 := json.Unmarshal(input, &pbAsset)
+		if err2 != nil {
+			return nil, errors.Wrap(err2, "unmarshaling to proto")
+		}
+		Log.Debug("getAssetFromJSON", "Unmarshal", "Unable to unmarshal with jsonpb but able to unmarshal with json")
```
katze120, line 139: Handling data conversion like this because the original approach using jsonpb does not handle [update_time field](https://github.com/googleapis/googleapis/blob/master/google/cloud/orgpolicy/v1/orgpolicy.proto#L290) of [Timestamp](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto) type. jsonpb expects Timestamp in ["{year}-{month}-{day}T{hour}:{min}:{sec}[.{frac_sec}]Z"](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto#L101) string format while CAI produces it in [the original format with nanos and seconds as integers](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/timestamp.proto#L127).

On the other hand, while `json.Unmarshal` can handle org policies, it cannot handle access policies with `oneof` field.

Hence trying both approach while converting.

morgante: Got it. I'd like to avoid putting critical code paths in an if statement though, instead could we refactor slightly to a separate function like this:

func unMarshallAsset() {
  err := jsonpb.Unmarshaler{AllowUnknownFields: true}
  if err == nil, return asset
  err = json.Unmarshal
  if err == nil, return asset
  return nil, err

cli/testdata/scorecard/policy-library/lib/constraints.rego
```diff
@@ -0,0 +1,36 @@
+#
```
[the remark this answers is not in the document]
katze120: why?
Not having lib is possible, but it means the same reusable codeblocks need to be copied in multiple test templates

cli/scorecard/violations.go
katze120: addressed in 1e1a678

cli/scorecard/violations.go
katze120: addressed in 1e1a678

cli/testdata/scorecard/policy-library/lib/constraints.rego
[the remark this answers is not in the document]
morgante: Essentially i don't want to have this repo be a source-of-truth for policy templates - we should just be copying them over from policy-library (which has the reusable code blocks and test infra to properly manage templates).

PR #685, opened 2020-05-13
Cloud Functions IAM member example

ocsig:


ocsig:
#591
https://github.com/GoogleCloudPlatform/deploymentmanager-samples/issues/494

PR #690, opened 2020-05-20
Add GitHub action for releasing CLI.

morgante:
This should help us to release faster with less toil.

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -7,6 +7,7 @@ on:
       - "cli/Makefile"
 env:
   CLI_BUCKET: gs://cft-cli
+  RELEASE_URL: https://api.github.com/repos/GoogleCloudPlatform/cloud-foundation-toolkit/releases/latest
 
 jobs:
   release-new-cli:
@@ -23,9 +24,8 @@ jobs:
 
       - name: Retrieve last released version
         run: |-
-          LAST_VERSION=$(curl --silent "https://api.github.com/repos/GoogleCloudPlatform/cloud-foundation-toolkit/releases/latest" | \
-            grep '"tag_name":' |
-            sed -E 's/.*"([^"]+)".*/\1/')
+          LAST_VERSION=$(curl --silent "${{env.RELEASE_URL}}" | \
+            jq --raw-output .tag_name)
           echo ::set-env name=LAST_VERSION::$LAST_VERSION
 
       - name: Build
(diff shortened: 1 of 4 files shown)


morgante:
@bharathkkb PTAL

.github/workflows/release-cli.yml
```diff
@@ -0,0 +1,98 @@
+name: Release new CLI versions
+on:
+  push:
+    branches: 
+      - "master"
+    paths:
+      - "cli/Makefile"
+env:
+  CLI_BUCKET: gs://cft-cli
```
bharathkkb, line 9: Are we using this in the Makefile? I also saw the same value being defined in the `cli/Makefile` as `BUCKET=gs://cft-cli`

.github/workflows/release-cli.yml
```diff
@@ -0,0 +1,98 @@
...
+      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+        with:
+          version: "286.0.0"
+          service_account_key: ${{ secrets.GCP_SA_KEY }}
+
+      - name: Retrieve last released version
+        run: |-
+          LAST_VERSION=$(curl --silent "https://api.github.com/repos/GoogleCloudPlatform/cloud-foundation-toolkit/releases/latest" | \
+            grep '"tag_name":' |
+            sed -E 's/.*"([^"]+)".*/\1/')
```
bharathkkb, line 28: Can we not use `jq` instead which maybe cleaner if all we need is the value of `tag_name` like we do for TF latest release here https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/1f47e570882acc69ee9f7ce6a9c6a12149d6f82e/.github/workflows/update-tooling.yml#L17

.github/workflows/release-cli.yml
morgante: Yeah, we have it set in both places. We could remove from Makefile eventually but for now I'm leaving as-is.

cli/cmd/version.go
```diff
@@ -17,6 +19,6 @@ var versionCmd = &cobra.Command{
 example: CFT CLI version v1.0.0`,
 	Args: cobra.NoArgs,
 	Run: func(cmd *cobra.Command, args []string) {
-		cmd.Println("CFT CLI version", Version)
+		fmt.Println(Version)
```
katze120, line 22: This change is not passing the [test case here](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/cli/cmd/version_test.go#L16)

morgante: Oops! Fixing: #701

PR #701, opened 2020-05-28
Fix output test cases broken in #690

morgante:


morgante:
> LGTM, cant we have make test as part of the CI pipeline before make release to catch these?

Yeah, it's a good idea with the caveat that `make test` requires access to GCP credentials.

bharathkkb:
@morgante arent we already authing using  `service_account_key: ${{ secrets.GCP_SA_KEY }}` in the release workflow?

morgante:
> @morgante arent we already authing using service_account_key: ${{ secrets.GCP_SA_KEY }} in the release workflow?

Good point! Yeah I think that should work, I'll plan to give it a try later.

PR #703, opened 2020-05-31
[KCC] Support Helm pubsub subscription solution

harshatech:
Adding Config Connector compatible YAML files to create a Pub/Sub subscription and Pub/Sub topic if it doesn't exist in your desired project and grant a specific member a role (default to roles/pubsub.viewer) for accessing the Pub/Sub subscription that just created.


maqiuyujoyce:
@AlexBulankou Why do we want to make all the fields, even set in the values.yaml, required? I thought the required fields are only the fields which requires customer input, i.e., the fields that are empty in the values.yaml.

BrianPi:
We should probably either [set Helm defaults](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) or make all values required, so that the Chart never breaks silently from changing the values YAML. It looks like some of the checked-in solutions don't follow that pattern right now.

maqiuyujoyce:
@BrianPi I think then we should set the defaults. It will be more consistent with what we do in kpt solutions.

AlexBulankou:
@maqiuyujoyce , @BrianPi , I agree with the suggestion, but I don't think we need to necessarily require that each solutions follows this completely, leaving these details up the solution author is ok.

PR #704, opened 2020-06-01
Scorecard - Org policy fix

katze120:
- Updated the approach to convert org policy data from json to proto and some explanation on why it needs to be handled that way

- Minor update to user guide to add Windows executable 

- Minor update to scorecard help menu to sync with latest features


katze120:
Update to use latest CV module to output ancestry path will be handled at later date. More planning and communication required due to downstream impact.

PR #734, opened 2020-06-30
Add trigger for example-foundation integration tests

bharathkkb:
Because `terraform-example-foundations` does not follow the usual module naming scheme `terraform-google-module` I had to make the trigger via a resource. Part of https://github.com/terraform-google-modules/terraform-example-foundation/issues/62

-  Created a folder `ci-example-foundation`  under the org

- Created int test trigger

/cc @rjerrems


infra/terraform/test-org/ci-triggers/triggers.tf
```diff
@@ -154,3 +154,28 @@ resource "google_cloudbuild_trigger" "terraform_example_foundation_lint" {
   filename = "build/cloudbuild.lint.yaml"
 }
 
+resource "google_folder" "ci-example-foundation" {
+  display_name = "ci-example-foundation"
+  parent       = "organizations/${local.org_id}"
```
bharathkkb, line 159: @morgante I am not sure if this should be under `folders/${ci-projects-id}`

morgante: @bharathkkb Please move it to the same location as all other tests. In fact, I'm not sure why this couldn't be treated the same as any other repo?

bharathkkb: @morgante wouldnt it would fail here https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/886d4c610e155a650a58444c0c7c2b90ba5a7cc3/infra/terraform/test-org/ci-triggers/locals.tf#L24
as the folder name will be `ci-example-foundation` and this transform would make it `terraform-google-example-foundation` which is invalid?

morgante: Good catch. Could we tweak that logic to allow adding repos with different names?

bharathkkb: We could have something like this https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/00c5a11276d01dc2ba569075871e5b44f2788eea/infra/terraform/test-org/ci-triggers/locals.tf#L18
but considering that we haven't cut any releases for `terraform-example-foundations` wouldn't it be better to rename to `terraform-google-example-foundations` to keep alignment?

bharathkkb: @morgante I have added an exception for `example-foundations` PTAL

PR #753, opened 2020-07-29
Fix error in dataproc schema

### LINKED ISSUE #752: Incorrect dataproc schema
The current dataproc schema says that `initializationActions` should have a `executableTimeout` property:

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/a9f6a74be09de47cc0e78dfd7efd748221223b8f/dm/templates/dataproc/dataproc.py.schema#L223-L231

but this is not correct. The dataproc template uses v1 of the `project.regions.clusters` API:

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/a9f6a74be09de47cc0e78dfd7efd748221223b8f/dm/templates/dataproc/dataproc.py#L124-L137

which actually expects a property named [`executionTimeout`](https://cloud.google.com/dataproc/docs/reference/rest/v1/ClusterConfig#NodeInitializationAction).

huy-nguyen:
Fixes #752.

### DIFF
--- a/dm/templates/dataproc/dataproc.py.schema
+++ b/dm/templates/dataproc/dataproc.py.schema
@@ -224,7 +224,7 @@ properties:
         executableFile:
           type: string
           description: The Cloud Storage URI of the executable file.
-        executableTimeout:
+        executionTimeout:
           type: string
           description: |
             The executable completion timeout, e.g. "3.5s". The default value


huy-nguyen:
@googlebot I signed it!

PR #773, opened 2020-08-23
feat: Add tf-bundle, parallel validate for cft-tools

bharathkkb:
Modules with lots of examples & fixtures (gke, example-foundations etc) can take +10 minutes to lint & validate via CB and +20 minutes locally, making it hard to smoke test. It also downloads providers per example/fixture often leading to dir sizes of +10G. 

This PR aims to 
- **reduce duplicate provider downloads using [terraform-bundle](https://github.com/hashicorp/terraform/tree/master/tools/terraform-bundle)**
    - if a [/test/bundle.hcl](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/21645e1769a849eff1543403ccaa9b4e10f2a45a/test/bundle.hcl) is found, it attempts to pre-download the required providers. If not it falls back to the default behavior.
    - gke time to validate locally _without bundle_: `24m35.551s`. dir size: `12.19GB`
    - gke time to validate locally _with bundle_: `10m4.602s`, dir size: `612MB`
    - not much of difference via CB, but it will help [prevent flaky runs](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/runs/1016534257) due to `registry.terraform.io` network errors.
- **enables validation test runs in parallel**
   - this is gated by requiring `ENABLE_PARALLEL` to be set otherwise defaults to usual
   - gke time to validate locally _no parallel_ with bundle: `10m4.602s`
   - gke time to validate locally _in parallel_ with bundle: `2m24.213s`
   - gke time to validate via CB _in parallel_ with bundle: `2 min 31 sec`. [Example run ](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/runs/1017109177)

TODO:

- [ ] TF version and image version


morgante:
@bharathkkb Good idea! I'm wondering if we could improve this by having a "default" bundle we include in the Docker image? Ideally I'd like to avoid having to include bundle config in each repo.

bharathkkb:
@morgante sounds good, I will add a default. And honestly this is only a major problem for a handful of repos, so if we just have bundles for those it should help with devex.

bharathkkb:
@morgante after some experimentation I think having a default bundle may adversely affect the time if the provider versions don't match. Across different modules, we pin to different provider versions and having one default does not make sense in that case. For instance if example foo pins to `"~> 3.16.0"` whereas our default bundle downloads `"~> 3.35.0"`, it will first unnecessarily download the bundle and still re download `"~> 3.16.0"` for foo. 

This effect will be compounded if we enable a default bundle across all our modules. Perhaps pessimistic pinning to a major release across all examples could be another solution. As a start, I propose individual bundles within example-foundations and gke and re evaluate if needed. Happy to hear additional thoughts/suggestions as well.

Long term there is also `terraform providers mirror`, currently in 0.13 which will bring additional functionality beyond the current bundler.

morgante:
@bharathkkb Thanks for investigating, I guess we start with bundles for the major/slowest modules. Please go ahead and ping me once this PR is ready for merge/review.

> Perhaps pessimistic pinning to a major release across all examples could be another solution

I want to avoid this so we don't unconsciously add dependencies on new provider versions without noticing.

bharathkkb:
@morgante this should be good to go. I will backport as `0.12.2` as well.
example-foundation: https://github.com/terraform-google-modules/terraform-example-foundation/pull/250
gke: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/641

bharathkkb:
@morgante gentle ping regarding this

infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -152,11 +164,22 @@ function check_terraform() {
   echo "Running terraform validate"
   # Change to a temporary directory to avoid re-initializing terraform init
   # over and over in the root of the repository.
-  find_files . -name "*.tf" -print \
+
+  # If enable parallel, run validate in parallel
```
morgante, line 168: @bharathkkb Is there a reason we wouldn't switch to parallel validation by default? (Fine to do this in a subsequent PR once we've tested it a bit.)

morgante (approved):

PR #832, opened 2020-11-20
fix: Add req for google-api-python-client for cleanup script

morgante:


bharathkkb (approved):
Is this utility for test project cleanup? My understanding was we were using [scheduled cleanup func](https://github.com/terraform-google-modules/terraform-google-scheduled-function/tree/f31b57ffd6dedf621fb1a2770a5e91ae0cf9fcd4/modules/project_cleanup).

morgante:
> Is this utility for test project cleanup? My understanding was we were using scheduled cleanup func.

No it's just a local util.

PR #852, opened 2020-12-15
feat: split out example-foundation int tests

bharathkkb:
Create separate example-foundation int triggers for each `example_foundation_int_test_modes` passed in via substitutions


morgante (approved):

bharathkkb:
This has been applied, will merge after testing.

PR #861, opened 2021-01-19
Update the 'outputs' section in gcs_bucket.py.schema

vinbs:
Correcting the format for the outputs section as per [1]

[1] https://cloud.google.com/deployment-manager/docs/configuration/expose-information-outputs#describing_outputs_in_schemas


morgante:
@ocsig Can you take a look here? It seems like all the other templates have `outputs.properties` so I don't think we should be changing only this template.

vinbs:
Will create a new pull request with fix for all the template schemas.

PR #885, opened 2021-03-15
Add Devtools light image

bharathkkb:
Adds a lighter version of devtools image with TF and gcloud


morgante (commented):
Could we update the full devtools image to use this new light image as a base, like we do with https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools-jenkins/Dockerfile?

bharathkkb:
@xingao267 @MartinPetkov PTAL for tooling. wget is already available and I have added jq and bash. Additionally also added git, ca-certificates and python3 (for gcloud).

bharathkkb:
@morgante I considered that but the base images for both are different. Full devtools image uses ruby-alpine for kitchen dependencies.

morgante:
@bharathkkb In that case, could we COPY the dependencies over from the base? Just so we can avoid having the install logic duplicated.

infra/build/developer-tools-light/README.md
```diff
@@ -0,0 +1,13 @@
+# Developer Tools Image Light
+
+This image is light version of developer-tools with Terraform and Cloud SDK.
```
morgante, line 3:
```suggestion
This image is a light version of developer-tools with Terraform and Cloud SDK.
```

morgante (approved):

MartinPetkov:
LGTM

infra/build/developer-tools-light/build/install_cloud_sdk.sh
```diff
@@ -0,0 +1,37 @@
+#! /bin/bash
```
morgante, line 1: @bharathkkb We shouldn't copy this file, instead either source directly from https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/build/install_cloud_sdk.sh or COPY the file over from the other image.

infra/build/developer-tools-light/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# limitations under the License.
+# Download and verify the integrity of the download first
+
+FROM sethvargo/hashicorp-installer:0.1.3 AS installer
+# Required to download and install Terraform
+ARG TERRAFORM_VERSION
+ENV TERRAFORM_VERSION ${TERRAFORM_VERSION}
+RUN /install-hashicorp-tool "terraform" "$TERRAFORM_VERSION"
+
+FROM alpine:3.12
```
xingao267, line 22: We probably should consider using one of the google's managed base image, e.g. ubuntu 18.04 (gcr.io/cloud-marketplace/google/ubuntu1804) as base for security purposes.

xingao267: ref: https://cloud.google.com/container-registry/docs/managed-base-images

morgante: Good idea, though note that Ubuntu would likely increase the image size substantially.

xingao267: We can use ubuntu16.04. It's used by most Cloud Build's published base builder images (e.g. https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/wget/Dockerfile#L1) and those images and its layers are cached in Cloud Build's VMs, so should be no time wasted for pulling the image even the base size is bigger than the alpine image.

xingao267: launcher.gcr.io/google/ubuntu16_04 or l.gcr.io/google/ubuntu16_04 was the old path for gcr.io/cloud-marketplace/google/ubuntu1604, we should use the cloud marketplace one instead.

xingao267: I just verified ubuntu1804 is also cached on Cloud Build VMs by default.

FETCHSOURCE
Fetching storage object: gs://xingao_cloudbuild/source/1615839015.59864-0a169fa6ce4e4e1a84ee18d02d97faf6.tgz#1615839015956184
Copying gs://xingao_cloudbuild/source/1615839015.59864-0a169fa6ce4e4e1a84ee18d02d97faf6.tgz#1615839015956184...
/ [1 files][  259.0 B/  259.0 B]                                                
Operation completed over 1 objects/259.0 B.                                      
BUILD
Pulling image: gcr.io/cloud-marketplace/google/ubuntu1804
Using default tag: latest
latest: Pulling from cloud-marketplace/google/ubuntu1804
9f2825a1003a: Already exists
7e79098fe9b4: Already exists
3c2cba919283: Already exists
Digest: sha256:d3d006e57a628ee5e03c5fbe6781baa8b8603c11463d44f1da06c92d75fcf343
Status: Downloaded newer image for gcr.io/cloud-marketplace/google/ubuntu1804:latest
gcr.io/cloud-marketplace/google/ubuntu1804:latest
test
PUSH
DONE

infra/build/developer-tools-light/build/install_cloud_sdk.sh
bharathkkb: So we cannot source directly since it is not available in the build context (different dir) and in the other image we [remove](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/2ddc9147cd8f94be3d64910b7dd1d737da96e6cb/infra/build/developer-tools/Dockerfile#L173) `/build`. A workaround could be in the makefile, we can copy it over at buildtime, but that would mean if a user tried to build from the directory directly `developer-tools-light` instead of makefile it will fail.

morgante: Should we move the build context to be in the parent directory instead then?

bharathkkb: @morgante done, the light image now builds at `/infra/build`. Note this still means if user runs docker build inside `developer-tools-light` they will run into an error.

infra/build/developer-tools-light/Dockerfile
bharathkkb: @xingao267 I have switched the base image. Some tools `ca-certificates, bash` etc are preinstall and I added `wget`.

xingao267: Thanks!

infra/build/developer-tools-light/README.md
```diff
@@ -0,0 +1,12 @@
+# Developer Tools Image Light
+
+This image is a light version of developer-tools with Terraform and Cloud SDK.
+
+## Building and Releasing
```
morgante, line 5: Perhaps we should cut this section down and simply say to run the Makefile in the parent directory?

morgante (approved):

infra/build/developer-tools-light/README.md
bharathkkb: done

bharathkkb:
Looks like actions has some degradation and our push event has not fired. It should automatically trigger, I'll update this when it runs.
https://twitter.com/githubstatus/status/1371561446463660035

bharathkkb:
@xingao267 @MartinPetkov the images should be available. New images will be pushed for any version bumps to gcloud/tf like https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/884.
```
gcr.io/cloud-foundation-cicd/cft/developer-tools-light:0.14.17
gcr.io/cloud-foundation-cicd/cft/developer-tools-light:0.14
gcr.io/cloud-foundation-cicd/cft/developer-tools-light:0
```

PR #913, opened 2021-05-10
fix: skip int tests if only doc changes

bharathkkb:
fixes https://github.com/terraform-google-modules/terraform-example-foundation/issues/468

Tested in example-foundation fork with 1 trigger:

https://github.com/bharathkkb/terraform-example-foundation/pull/3 skipped as only md changes

https://github.com/bharathkkb/terraform-example-foundation/pull/4 ran as expected


bharathkkb:
this has been applied

PR #943, opened 2021-06-11
Use Alpine base for light developer tools container

xingao267:
It actually has fewer vulnerabilities as discovered by GCR
Vulnerability scanning. It's also smaller.


xingao267:
The `gcr.io/cloud-marketplace/google/ubuntu1804:latest` based has 54 vulnerability and the alpine based has none.

infra/build/developer-tools-light/Dockerfile
```diff
@@ -19,8 +19,8 @@ ARG TERRAFORM_VERSION
 ENV TERRAFORM_VERSION ${TERRAFORM_VERSION}
 RUN /install-hashicorp-tool "terraform" "$TERRAFORM_VERSION"
 
-FROM gcr.io/cloud-marketplace/google/ubuntu1804:latest
-RUN apt-get update -y && apt-get install -y -q \
+FROM alpine
```
bharathkkb, line 22: let's pin this image to a tag?

infra/build/developer-tools-light/Dockerfile
```diff
@@ -29,6 +29,9 @@ RUN apt-get update -y && apt-get install -y -q \
 # Install Terraform
 COPY --from=installer /software/terraform /usr/local/bin/terraform
 
+# bash is preferred over /bin/sh
+RUN apk add --no-cache bash parallel
```
bharathkkb, line 33: Why do we need a separate step, we do an apk add above and these can be appended there?

xingao267: Fixed

infra/build/developer-tools-light/Dockerfile
xingao267: Done

bharathkkb (approved):

bharathkkb:
@xingao267 do you want to do a patch version bump to preserve the `1.0.0` image with ubuntu or would you like this new image to be tagged `1.0.0`?
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/22dc2cc6ec5bead3204b4a6e0ce6ebafa4a864f7/infra/build/Makefile#L31

xingao267:
> @xingao267 do you want to do a patch version bump to preserve the `1.0.0` image with ubuntu or would you like this new image to be tagged `1.0.0`?
> https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/22dc2cc6ec5bead3204b4a6e0ce6ebafa4a864f7/infra/build/Makefile#L31

Updated the version. Let's keep the existing 1.0.0 tag.

PR #954, opened 2021-07-08
feature: add support for addlicense based header check

### LINKED ISSUE #926: The license header templates should allows headers created by the addlicense tool
#### Issue
- The `check_headers` step on the linters uses a set of [template files](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/tree/master/infra/build/developer-tools/build/verify_boilerplate/boilerplate) to validate if the license headers are valid 
- However, the template files used here are quiet different to the ones generated by the [addlicense tool](https://github.com/google/addlicense) used in many places. 
- As a result for projects that rely on the `addlicense` tool to generate the headers, the linter here fails when checked agaisnt these samples.

#### Fix
- Align the templates here to match [those used](https://github.com/google/addlicense/tree/master/testdata/expected) by the addlicense tool 
- Ensure that the current templates are also kept valid in order to ensure existing modules that use these templates don't start breaking

Shabirmean:
#### Changes
- Added script to install the [addlicense](https://github.com/google/addlicense) tool inside the images
- Updated the `check_headers` method to include check using the addlicense tool
- The header check is deemed failure only if it's caught by both tools - _the legacy_ and the _addlicense tool_. If one of them catches it and the test passes for the other - we deem it as a valid file. This is ensure test cases based on the legacy method doesn't break
- There is a new environment variable `EXCLUDE_HEADER_CHECK` that is introduced to take in files to be skipped by the addlicense tool


infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -384,7 +384,22 @@ function check_headers() {
   echo "Checking file headers"
   # Use the exclusion behavior of find_files
   find_files . -type f -print0 \
-    | compat_xargs -0 python /usr/local/verify_boilerplate/verify_boilerplate.py
+    | compat_xargs -0 python /usr/local/verify_boilerplate/verify_boilerplate.py 2>&1 \
+    | grep -v 'have incorrect boilerplate headers' > legacy_check.log
+  if [[ -v EXCLUDE_HEADER_CHECK ]]; then
```
Shabirmean, line 389: I am not sure if I have to introduce **_EXCLUDE_HEADER_CHECK_** to other places or do any other updates to make it working like with [EXCLUDE_LINT_DIRS](https://github.com/terraform-google-modules/terraform-example-foundation/blob/1fad10b2af93069a7369cd100a3fcf2cddaa6f74/build/lint.cloudbuild.yaml#L23)

morgante: Do we need to add a separate exclude for this? I'd prefer to continue using `EXCLUDE_LINT_DIRS`.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -384,7 +384,22 @@ function check_headers() {
...
+    SKIP_STRING=""
+    for file in ${EXCLUDE_HEADER_CHECK}
+    do
+      SKIP_STRING="$SKIP_STRING -skip $file"
+    done
+    stored_cmd="addlicense -check $SKIP_STRING ./ 2>&1 | grep -v 'skipping this file'"
+    eval "$stored_cmd" > new_check.log
+  else
+    addlicense -check ./ 2>&1 > new_check.log
```
morgante, line 398: We should not check globally. Instead, we should use `find_files` to find the directories to check (like the legacy check does) and then run the check against each.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -384,7 +384,22 @@ function check_headers() {
...
+  fi
+  # list only the files caught by both methods, if not assume success header check
```
morgante, line 400: What's the reason for checking both? Are they not compatible?

morgante (changes requested):
Thanks for working on this.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
Shabirmean: I'm not sure how its (`EXCLUDE_LINT_DIRS`) expected usage is. My thinking was that with `EXCLUDE_LINT_DIRS ` we cannot exclude specific files - but only whole directories. Also can we provide a list of comma separated options? And the other thing is, what if we wanted to only skip `header_checks` and not all the other linting. These were the things I was taking into consideration. WDYT?

morgante: Correct, `EXCLUDE_LINT_DIRS` only works for directories. So far, that has been sufficient.

I don't think we have any cases of only wanting to exclude a single file from header checks. Given that, I suggest we don't include it unless/until we have a specific use case.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
Shabirmean: Yes, as discussed in this [thread](https://chat.google.com/room/AAAAWaRnhC8/23oJnRQ8Zv8) there are cases where the `legacy_check` fails when the `addlicense_check` passes. It might be the case the other way round too. So as guided in [the thread](https://chat.google.com/room/AAAAWaRnhC8/23oJnRQ8Zv8) I am trying to make sure, previous CI checks that were passing with just the `legacy_check` continue to pass. 

Another way would be to advertise migration to addlicense and ask everyone to fix the headers by runnign the addlicense tool going forward.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
Shabirmean: So I am working on [this PR](https://github.com/GoogleCloudPlatform/anthos-samples/pull/59) which has a `cloud-config` file to be passed into the cloud-init system. If this file has any headers the cloud-init fails. I have to just skip only this file. So this is the case I am trying to address here. 

For now, I have to evade it by creating a directory to only hold this file and point to it in `EXCLUDE_LINT_DIRS`. Ideally, I wouldn't want to do this and I want the other linting to run on that file. Do you think this is a valid use case?

morgante: Got it. Since you have a use case, I'm fine to include it.

However, I do suggest we use a generalized approach with find_files instead of relying on specific exclusion logic for `addlicense` specifically, particularly because the skip argument seems to be [for extensions](https://github.com/google/addlicense/blob/master/main.go#L77).

infra/build/developer-tools/build/scripts/task_helper_functions.sh
morgante: Got it. Can you have it log cases where addlicense fails with a warning? I'd like to diagnose those / phase out the legacy check over time.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
Shabirmean: Thank you. 

Yes, looking into what `find_files` does. The `addlicense` skip is not properly documented. It works for specific files too. I tested it out.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
Shabirmean: Sure. I will add the logs for this

bharathkkb: @Shabirmean I think it could be useful to have a `fix_headers` fn exposed which targets the same files to help ease the migration and ongoing maintenance.

Shabirmean: @bharathkkb - sure. I can add that.

@morgante - what is logs you expect?
- files that pass legacy check but fail the addlicense one? *(this I haven't encountered)*
- files that pass the addlicense but fail the legacy check? *(this is the one that I have encountered)*

infra/build/developer-tools/build/scripts/task_helper_functions.sh
Shabirmean: This has been fixed. I have used the `find_files` utility function to get the list of files to check

infra/build/developer-tools/build/scripts/task_helper_functions.sh
Shabirmean: Resolving this thread. Please re-open if need be.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
morgante: @Shabirmean I was hoping for logs of "files that pass legacy check but fail the addlicense one."

If you haven't found this so far, I think we should just remove the legacy check entirely. It sounds like `addlicense` is backwards-compatible with our existing headers.

infra/build/developer-tools/build/scripts/task_helper_functions.sh
morgante: To be clear, I was hoping we would move the `EXCLUDE_HEADER_CHECK` logic into `find_files` so it's consistent (instead of relying on specific `-skip` flag in `addlicense`. It's a minor point but I'd rather have our exclusion logic centralized and tool-agnostic.

Shabirmean: All exclusion logic moved to find_files in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/954/commits/268a08357246a0794b0b24ac274a48c400a4874f

infra/build/developer-tools/build/scripts/task_helper_functions.sh
Shabirmean: Added new method called `fix_headers`
Moved exclusion logic to `find_files`
Removed all legacy header checks and kept only `addlicence`

infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -391,12 +406,27 @@ function prepare_test_variables() {
...
-    | compat_xargs -0 python /usr/local/verify_boilerplate/verify_boilerplate.py
+  # Use the exclusion behavior of find_files(); a second argument
+  # "for_header_check" is passed in, to ensure filtering based on the evironment
+  # variable EXCLUDE_HEADER_CHECK is done only when find_files is called here
+  find_files . for_header_check -type f -print0 | compat_xargs -0 addlicense -check 2>&1
+}
+
+# Add license headers to the files in the project. If a list of files are provided
+# as an input argument then those files are updated to have the license header.
+# If not find_files() funciton is used to get the list of applicable files from
```
morgante, line 417:
```suggestion
# If not find_files() function is used to get the list of applicable files from
```

morgante (approved):
Thanks for working on this.

PR #956, opened 2021-07-11
fix: fmt lint with cache dir

### LINKED ISSUE #955: devtools: terraform fmt linter not working in 1.0
It looks like #936 broke the `terraform fmt` linter. When running the linter for the first time, this error is raised:

```
Running terraform fmt
There are some problems with the CLI configuration:
╷
│ Error: The specified plugin cache dir /workspace/test/integration/tmp/.terraform cannot be opened: stat /workspace/test/integration/tmp/.terraform: no such file or directory
│
╵

As a result of the above problems, Terraform may not behave as intended.
```

Note this only happens the *first* time. Once the cache has been built, tests start passing.

bharathkkb:
fixes #955

### DIFF
--- a/infra/build/developer-tools/build/scripts/task_helper_functions.sh
+++ b/infra/build/developer-tools/build/scripts/task_helper_functions.sh
@@ -135,6 +135,11 @@ function check_terraform() {
   local rval rc
   rval=0
 
+  # if TF_PLUGIN_CACHE_DIR is set, create TF_PLUGIN_CACHE_DIR
+  if [[ ! -z "${TF_PLUGIN_CACHE_DIR}" ]]; then
+    mkdir -p ${TF_PLUGIN_CACHE_DIR}
+  fi
+
   # fmt is before validate for faster feedback, validate requires terraform
   # init which takes time.
   echo "Running terraform fmt"


morgante (approved):
I recommend we also remove the `mkdir` from the Dockerfile: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/Dockerfile#L168

It's confusing to have it in both, especially since the Dockerfile creation doesn't have any impact (since the /workspace volume overwrites it).

bharathkkb:
@morgante without mkdir in the dockerfile, if a volume is not mounted as `/workspace` TF will throw an error. This will also happen at build time since do a [terraform version](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/96cd1248762c4e235b580bc9d96314b08a5aebe4/infra/build/developer-tools/Dockerfile#L171)
```
terraform version
There are some problems with the CLI configuration:
╷
│ Error: The specified plugin cache dir /workspace/test/integration/tmp/.terraform cannot be opened: stat /workspace/test/integration/tmp/.terraform: no such file or directory
```

morgante:
> @morgante without mkdir in the dockerfile, if a volume is not mounted as `/workspace` TF will throw an error. This will also happen at build time since do a [terraform version](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/96cd1248762c4e235b580bc9d96314b08a5aebe4/infra/build/developer-tools/Dockerfile#L171)
> 
> ```
> terraform version
> There are some problems with the CLI configuration:
> ╷
> │ Error: The specified plugin cache dir /workspace/test/integration/tmp/.terraform cannot be opened: stat /workspace/test/integration/tmp/.terraform: no such file or directory
> ```

Got it. In real usage we mount a volume (either locally or in Cloud Build), but I missed that we also use it in the build itself.

PR #977, opened 2021-08-19
feat: initialize KRM blueprint testing

bharathkkb:


infra/blueprint-test/.gitignore
```diff
@@ -67,3 +67,5 @@ credentials.json
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+# Temp dir
+**/.build/*
```
morgante, line 71: Nit: newline please.

infra/blueprint-test/examples/krm_blueprints_with_test/test/integration/vpc-example-custom/vpc_custom_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+func TestVPCCustomBlueprint(t *testing.T) {
+	networkBlueprint := krmt.NewKRMBlueprintTest(t,
+		krmt.WithUpdateCommit("2b93fd6d4f1a3eabdf4dfce05b93ccb1f9f671c5"),
+	)
+	networkBlueprint.DefineVerify(
+		func(assert *assert.Assertions) {
+			networkBlueprint.DefaultVerify(assert)
+			op := gcloud.Run(t, fmt.Sprintf("compute networks describe custom-network --project %s", networkBlueprint.ValFromEnv("PROJECT_ID")))
+			assert.Equal("GLOBAL", op.Get("routingConfig.routingMode").String(), "should have be GLOBAL")
```
morgante, line 20:
```suggestion
			assert.Equal("GLOBAL", op.Get("routingConfig.routingMode").String(), "should be GLOBAL")
```

infra/blueprint-test/pkg/git/git.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/infra/blueprint-test/pkg/utils"
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/shell"
+	"github.com/mitchellh/go-testing-interface"
+)
+
+type CmdCfg struct {
+	gitBinary   string         // git binary
+	dir         string         // dir to execute commands in
+	commitFlags []string       // additional flags when running git commit
```
morgante, line 16: Since these are specific to the CommitWithMsg method, this should be moved to simply being a parameter of that function.

infra/blueprint-test/pkg/kpt/kpt.go
```diff
@@ -0,0 +1,69 @@
...
+type CmdCfg struct {
+	kptBinary string         // kpt binary
+	dir       string         // dir to execute commands in
+	logger    *logger.Logger // custom logger
+	t         testing.TB     // TestingT or TestingB
+}
+
+type cmdOption func(*CmdCfg)
+
+func WithDir(dir string) cmdOption {
```
morgante, line 19: Small note, but since it looks like the `dir`, `logger`, and `binary` functions are getting fairly repetitive across these utils, we could consolidate them into a shared, embedded `BinaryCfg` struct: https://golang.org/doc/effective_go#embedding

infra/blueprint-test/pkg/kpt/setters.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/GoogleContainerTools/kpt-functions-catalog/functions/go/list-setters/listsetters"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// UpsertSetters inserts or updates setters if apply-setters fn config is discovered.
+func UpsertSetters(nodes []*yaml.RNode, setters map[string]string) error {
+	kf, err := listsetters.FindKptfile(nodes)
```
morgante, line 16: Does this mean we only act on the first Kptfile found (presumably the root)?

infra/blueprint-test/pkg/krmt/krm.go
```diff
@@ -0,0 +1,312 @@
...
+	"github.com/otiai10/copy"
+	"github.com/stretchr/testify/assert"
+)
+
+const tmpBuildDir = ".build"
+
+var CommonSetters = []string{"PROJECT_ID", "BILLING_ACCOUNT_ID", "ORG_ID"}
+
+// KRMBlueprintTest implements bpt.Blueprint and stores information associated with a KRM blueprint test.
+type KRMBlueprintTest struct {
```
morgante, line 26: A lot of this could be combined with the Terraform version (ex. name, logger, etc.). Do you plan to consolidate in a later PR?

infra/blueprint-test/pkg/krmt/krm.go
```diff
@@ -0,0 +1,312 @@
...
+			b.logger.Logf(b.t, "Setting env var %s as setter %s: %s", s, sKey, sVal)
+			setters[sKey] = sVal
+		}
+	}
+	// merge user provided setters with env discovered setters
+	// user provided setters override env discovered setters
+	b.setters = kpt.MergeSetters(setters, b.setters)
+}
+
+// ValFromEnv returns value for a given env var, fails test if env var is not set.
```
morgante, line 170: What's the reason for pulling this into here vs. leaving it as a common util?

infra/blueprint-test/pkg/krmt/krm.go
```diff
@@ -0,0 +1,312 @@
...
+func (b *KRMBlueprintTest) ValFromEnv(k string) string {
+	v, err := utils.ValFromEnv(k)
+	if err != nil {
+		b.t.Fatal(err)
+	}
+	return v
+}
+
+// setupBuildDir prepares build dir with configs from bpDir.
```
morgante, line 179: Isn't this actually configs from example/fixture directory (not the blueprint itself)?

infra/blueprint-test/pkg/krmt/krm.go
```diff
@@ -0,0 +1,312 @@
...
+// DefaultApply installs resource-group, initializes inventory and apply pkg.
+func (b *KRMBlueprintTest) DefaultApply(assert *assert.Assertions) {
+	b.kpt.RunCmd("live", "install-resource-group")
+	b.kpt.RunCmd("live", "init")
+	b.kpt.RunCmd("live", "apply")
+}
+
+// DefaultVerify polls resource statuses until current.
+func (b *KRMBlueprintTest) DefaultVerify(assert *assert.Assertions) {
+	b.kpt.RunCmd("live", "status", "--output", "json", "--poll-until", "current", "--timeout", b.timeout)
```
morgante, line 242: This should probably be moved into the apply behavior, since waiting for reconciliation is more akin to what happens during `terraform apply` (and I anticipate every test having to do this).

infra/blueprint-test/pkg/krmt/krm.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// DefaultTeardown destroys resources from cluster and polls until deleted.
+func (b *KRMBlueprintTest) DefaultTeardown(assert *assert.Assertions) {
+	b.kpt.RunCmd("live", "destroy")
+	b.kpt.RunCmd("live", "status", "--output", "json", "--poll-until", "deleted", "--timeout", b.timeout)
+}
+
+// AutoDiscoverAndTest discovers KRM config from examples/fixtures and runs tests.
+func AutoDiscoverAndTest(t *gotest.T) {
```
morgante, line 252: This should probably go in discovery.

morgante (changes requested):
Looking good, nice to see how similar KRM and Terraform are.

infra/blueprint-test/.gitignore
bharathkkb: done

infra/blueprint-test/examples/krm_blueprints_with_test/test/integration/vpc-example-custom/vpc_custom_test.go
bharathkkb: done

infra/blueprint-test/pkg/kpt/kpt.go
bharathkkb: I was planning to refactor these and TF/KRMBlueprintTests to use this pattern in a follow up PR

infra/blueprint-test/pkg/kpt/setters.go
bharathkkb: Correct, `FindKptfile` looks for a PathAnnotation that is `Kptfile` which will only match the root kptfile and not any other Kptfiles. My thought was as `render` happens bottoms up, upserting a setter at the root will affect nested pkgs that have the same setters too. In the future if we want to have a more targeted upsert for nested pkgs we can add that as a separate helper.

infra/blueprint-test/pkg/krmt/krm.go
bharathkkb: Yup, along with the binary helpers too.

infra/blueprint-test/pkg/krmt/krm.go
bharathkkb: In go/blueprints-e2e there was a suggestion to put this into the pkg so we don't have to keep plumbing `t`

infra/blueprint-test/pkg/krmt/krm.go
bharathkkb: Yeah its the configs in the example dir. This name was misleading so I switched it to `exampleDir`.

infra/blueprint-test/pkg/krmt/krm.go
bharathkkb: Makes sense moved this to apply and added a helper to mimic default TF verify behavior (i.e assert that no resource changes after apply)

infra/blueprint-test/pkg/krmt/krm.go
bharathkkb: Currently TF has this in tft and was planning on consolidating both into `discovery` in a follow up PR once we have a shared embedded struct.

infra/blueprint-test/pkg/kpt/kpt.go
morgante: Ack, sounds good.

infra/blueprint-test/pkg/kpt/setters.go
morgante: Sounds right, I think we only need to act on the root.

infra/blueprint-test/pkg/krmt/krm.go
morgante: I see, I probably suggest that but looking at it now I think I might prefer a common util. It's fine this way though, so your decision.

infra/blueprint-test/pkg/krmt/krm.go
morgante: Sounds good.

morgante (approved):

infra/blueprint-test/pkg/git/git.go
bharathkkb: done

infra/blueprint-test/pkg/krmt/krm.go
bharathkkb: I moved this to common utils, we can add a helper in krmt in the future if there is friction.

PR #982, opened 2021-08-26
feat: add KRM benchmark example

bharathkkb:
This adds a KRM benchmark example

- Also moves `pkg init` and `install-resource-group` from DefaultInit to DefaultApply for krmt. 
Reasoning for this is because `DefaultApply` maybe run multiple times. In particular `pkg init` is not idempotent and will error out if a pkg is already inited.


morgante:
> @morgante We are trying to benchmark KCC actuation time from object creation to resource actuation. In general across tools, we are trying to benchmark time from when a tool is invoked create N resources to when those resources are actually created.

Right, what I'm confused on is are we trying to measure `kpt` or `KCC` here? We can chat offline.

infra/blueprint-test/test/krm_simple_pubsub_bench_test.go
```diff
@@ -0,0 +1,64 @@
...
+			for _, name := range topicNames {
+				benchmark.CreateVariant(b, blueprintDir, buildDir, name, map[string]string{"topic-name": name, "project-id": projectID})
+			}
+			// render variants
+			// TODO(bharathkkb): this is currently done in serial by kpt and can be slow for bigger topicCounts
+			// We should look into doing this in parallel possibly bundling variant creation with rendering
+			kptHelper.RunCmd("fn", "render")
+			kptHelper.RunCmd("live", "install-resource-group")
+			kptHelper.RunCmd("live", "init")
+			pubsubTest := krmt.NewKRMBlueprintTest(b, krmt.WithDir(buildDir), krmt.WithBuildDir(buildDir), krmt.WithUpdatePkgs(false))
```
morgante, line 48: I think we should move this into `benchmark` so I just call `benchmark.CreateTestVariant` which handles creating the variant directory and an associated test.

morgante: @bharathkkb Thoughts on this?

bharathkkb: @morgante done

morgante (approved):

PR #986, opened 2021-09-07
feat: add test yaml config

bharathkkb:
This implements parsing additional test configuration in a KRM format. Currently the only feature supported is to specify tests to skip.


infra/blueprint-test/pkg/discovery/discover.go
```diff
@@ -69,21 +69,31 @@ func FindTestConfigs(t testing.TB, intTestDir string) []string {
 	if err != nil {
 		t.Logf("Error discovering examples: %v", err)
 	}
-	testCases := make([]string, 0)
-	//TODO(bharathkkb): add overrides
+	testCases := make(map[string]string)
+	testConfig, err := getTestConfig(defaultTestConfigFilename)
```
morgante, line 73: I was thinking each blueprint test directory would have its own config. What do you think of that (ex. `examples/my-test/test.yaml`)?

Especially for mono-repos. that seems better than collapsing all config into one file.

bharathkkb: That sounds good to me. Does this API per test seem reasonable to start with?
```yaml
apiVersion: blueprints.cloud.google.com/v1alpha1
kind: BlueprintTest
metadata:
  name: test
spec:
  skip: true
```

morgante: Yep, LGTM.

bharathkkb: done

infra/blueprint-test/pkg/discovery/config.go
```diff
@@ -0,0 +1,64 @@
...
+type BlueprintTestConfig struct {
+	yaml.ResourceMeta `json:",inline" yaml:",inline"`
+	Spec              struct {
+		Skip bool `json:"skip" yaml:"skip"`
+	} `json:"spec" yaml:"spec"`
+	Path string
+}
+
+// ShouldSkipTest checks if a test should be skipped
+func (b BlueprintTestConfig) ShouldSkipTest() bool {
```
morgante, line 25: This should be a method on BlueprintTest, not BlueprintTestConfig. The fact that it comes from yaml is an implementation detail.

morgante (changes requested):

infra/blueprint-test/pkg/discovery/config.go
bharathkkb: added to respective BlueprintTests

infra/blueprint-test/pkg/discovery/config.go
```diff
@@ -0,0 +1,64 @@
...
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const (
+	DefaultTestConfigFilename = "test.yaml"
+	blueprintTestKind         = "BlueprintTest"
+	blueprintTestAPIVersion   = "blueprints.cloud.google.com/v1alpha1"
+)
+
+type BlueprintTestConfig struct {
```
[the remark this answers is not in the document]
bharathkkb: @morgante we haven't added the `BlueprintTest` shared struct yet. I was planning to refactor that after https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/987 during which I can investigate embedding `BlueprintTestConfig` into `BlueprintTest`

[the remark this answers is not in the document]
morgante: Ack, this can be done as part of that.

infra/blueprint-test/pkg/krmt/krm.go
```diff
@@ -24,20 +24,21 @@ var CommonSetters = []string{"PROJECT_ID", "BILLING_ACCOUNT_ID", "ORG_ID"}
...
-	updateCommit string                   // specific commit to update to
-	logger       *logger.Logger           // custom logger
-	t            testing.TB               // TestingT or TestingB
-	init         func(*assert.Assertions) // init function
-	apply        func(*assert.Assertions) // apply function
-	verify       func(*assert.Assertions) // verify function
-	teardown     func(*assert.Assertions) // teardown function
+	name         string                         // descriptive name for the test
+	exampleDir   string                         // directory containing KRM blueprint example
+	testConfig   *discovery.BlueprintTestConfig // additional blueprint test configs
```
morgante, line 29: This should be embedded, not pointed to.

morgante (changes requested):

infra/blueprint-test/pkg/krmt/krm.go
bharathkkb: done

morgante (approved):

PR #987, opened 2021-09-07
feat: refactor binary interactions in test framework

bharathkkb:
Consolidates test framework interaction with binaries into a binary pkg

### DIFF
--- a/infra/blueprint-test/pkg/binary/bin.go
+++ b/infra/blueprint-test/pkg/binary/bin.go
@@ -1,0 +1,58 @@
+package binary
+
+import (
+	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/infra/blueprint-test/pkg/utils"
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/mitchellh/go-testing-interface"
+)
+
+type BinaryCfg struct {
+	binary string         // path to  binary
+	logger *logger.Logger // custom logger
+	dir    string         // dir to execute commands in
+}
+
+type BinaryOption func(*BinaryCfg)
+
+func WithBinary(binary string) BinaryOption {
+	return func(f *BinaryCfg) {
+		f.binary = binary
+	}
+}
+
+func WithLogger(logger *logger.Logger) BinaryOption {
+	return func(f *BinaryCfg) {
+		f.logger = logger
+	}
+}
+
+func WithDir(dir string) BinaryOption {
+	return func(f *BinaryCfg) {
+		f.dir = dir
+	}
+}
+
+// NewBinaryConfig sets defaults and validates configuration for running a binary
+func NewBinaryConfig(t testing.TB, bin string, opts ...BinaryOption) *BinaryCfg {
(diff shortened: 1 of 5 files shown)


infra/blueprint-test/pkg/gcloud/gcloud.go
```diff
@@ -20,24 +20,23 @@ package gcloud
...
 	"github.com/gruntwork-io/terratest/modules/shell"
 	"github.com/mitchellh/go-testing-interface"
 	"github.com/tidwall/gjson"
 )
 
 type CmdCfg struct {
-	gcloudBinary string         // path to gcloud binary
-	commonArgs   []string       // common arguments to pass to gcloud calls
-	logger       *logger.Logger // custom logger
+	*binary.BinaryCfg                       // binary config
```
morgante, line 30: What's the reason to make this a pointer vs embedding the struct entirely?

infra/blueprint-test/pkg/binary/bin.go
```diff
@@ -0,0 +1,58 @@
...
+		opt(b)
+	}
+	err := utils.BinaryInPath(b.binary)
+	if err != nil {
+		t.Fatalf("unable to find binary %s in path: %v", b.binary, err)
+	}
+	return b
+}
+
+func (b *BinaryCfg) GetBinary() string {
```
morgante, line 48: We probably shouldn't use getters/setters. If these properties should be public, simply export them.

infra/blueprint-test/pkg/gcloud/gcloud.go
```diff
@@ -20,24 +20,23 @@ package gcloud
...
+	bOpts             []binary.BinaryOption // binary options
+	commonArgs        []string              // common arguments to pass to gcloud calls
 }
 
 type cmdOption func(*CmdCfg)
 
-func WithBinary(gcloudBinary string) cmdOption {
+func WithBinaryOptions(bOpts ...binary.BinaryOption) cmdOption {
```
morgante, line 37: We shouldn't need this layer of indirection. As a consumer of the `gcloud` package, I shouldn't need to know if options are common (across all binaries) or specific to gcloud.

morgante (changes requested):

infra/blueprint-test/pkg/gcloud/gcloud.go
bharathkkb: @morgante in that case should we continue to keep the `WithBinary` and other options within the `gcloud` pkg?

morgante: @bharathkkb No, I think it should still be possible to share the options by using a interfaces. If this doesn't work I can give it a try.

bharathkkb: Thanks, I'll try that approach

morgante:
@bharathkkb What's the status of this PR?

bharathkkb:
@morgante I de prioritized this for some feature work. I will pick this up soon.

PR #995, opened 2021-09-27
feat: test foundation reconciler

bharathkkb:
This adds a scheduler job to trigger a build that will reconcile test foundations to the known good state by triggering downstream builds for each env. Also removes old config for generating SA keys which we will replace with WIF.


morgante (approved):

bharathkkb:
This has been applied

PR #997, opened 2021-09-29
feat: support complex type setup outputs

bharathkkb:
This adds support for setting complex type outputs from setup as inputs. 

TF requires list/map env vars to be [formatted as an HCL expression](https://www.terraform.io/docs/cli/commands/plan.html#input-variables-on-the-command-line) but Terratest does not directly expose its [conversion function](https://github.com/nlowe/terratest/blob/2ba8cabb767dd7ab1a918138540501fee8d6dd56/modules/terraform/format.go#L66).  As  a workaround, we extract the formatted HCL expression from the var arg slice. We cannot switch from the env var approach to var args directly due to https://github.com/hashicorp/terraform/issues/28764

### DIFF
--- a/infra/blueprint-test/pkg/tft/terraform_test.go
+++ b/infra/blueprint-test/pkg/tft/terraform_test.go
@@ -28,7 +28,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestTFBlueprintTest_getTFOutputsAsInputs(t *testing.T) {
+func TestGetTFOutputsAsInputs(t *testing.T) {
 	tests := []struct {
 		name  string
 		input string
--- a/infra/blueprint-test/pkg/tft/terraform_test.go
+++ b/infra/blueprint-test/pkg/tft/terraform_test.go
@@ -95,7 +95,7 @@ func getTFOutputMap(t *testing.T, tf string) map[string]interface{} {
 	return terraform.OutputAll(t, tOpts)
 }
 
-func Test_getKVFromOutputString(t *testing.T) {
+func TestGetKVFromOutputString(t *testing.T) {
 	tests := []struct {
 		name    string
 		kv      string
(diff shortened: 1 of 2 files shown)


morgante (changes requested):
Since this is using several libraries can you add a test?

bharathkkb:
@morgante added a few cases for this feature

infra/blueprint-test/pkg/tft/terraform_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"path"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	testingiface "github.com/mitchellh/go-testing-interface"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTFBlueprintTest_getTFOutputsAsInputs(t *testing.T) {
```
morgante, line 31:
```suggestion
func TestGetTFOutputsAsInputs(t *testing.T) {
```

morgante (approved):

infra/blueprint-test/pkg/tft/terraform_test.go
```diff
@@ -0,0 +1,124 @@
...
+	assert.NoError(err)
+
+	// apply tf and get outputs
+	tOpts := &terraform.Options{TerraformDir: path.Dir(tfFilePath), Logger: logger.Discard}
+	terraform.Init(t, tOpts)
+	terraform.Apply(t, tOpts)
+	return terraform.OutputAll(t, tOpts)
+}
+
+func Test_getKVFromOutputString(t *testing.T) {
```
morgante, line 98:
```suggestion
func TestGetKVFromOutputString(t *testing.T) {
```

bharathkkb: This seems to be a thing in gotests when it generates test boilerplate https://github.com/golang/go/issues/33688. We probably wont hit that described edgecase.

morgante: Interesting. I tend to agree with the conclusion here:
> Notice that even within this arguably easy-to-parse version, I tend to elide the "T_" prefix when I think it doesn't cause confusion, i.e. if there's only one major exported type in the package.

That's definitely the case here.

PR #1008, opened 2021-10-19
feat: setup WIF for devtools image push

bharathkkb:
This setups up WIF for use with [google-github-actions/auth](https://github.com/google-github-actions/auth) to push devtools image


bharathkkb:
This has been applied

PR #1013, opened 2021-10-27
A workaround for illegal base64 data at input

hussainak:
A workaround for illegal base64 data at input (that contains hyphens and underscores). This is happening in the iam_policy.json CAI file that has etag with hyphens and underscores.


morgante:
Superseded by #1017.

hussainak:
@googlebot I consent

PR #1017, opened 2021-11-10
A workaround for illegal base64 data at input

morgante:
This supersedes #1013, with added tests.


hussainak:
@googlebot I consent

PR #1019, opened 2021-11-24
Add magic_modules service account to project perm list

ScottSuarez:
In order to run tests during magic modules ci we need to add it's service account to the permissions list here.


morgante (approved):
This has been applied.

PR #1022, opened 2021-12-01
correct cloud build service account for magic modules ci

ScottSuarez:
Correcting a mistake in my last PR where I used added incorrect cloudbuild SA


morgante (approved):
This has been applied.

PR #1025, opened 2021-12-07
fix(bptest): compute config discovery from int test dir

bharathkkb:
`GetConfigDirFromTestDir` is used by the framework to compute config dir based on the test dir name. Currently it does not take into account test dir path as example/fixtures are well known. However for use in CLI, I would like to compute config dir from relative to a test dir path.

### DIFF
--- a/infra/blueprint-test/pkg/discovery/discover_test.go
+++ b/infra/blueprint-test/pkg/discovery/discover_test.go
@@ -27,7 +27,7 @@ import (
 const (
 	baseDir     = "testdata/tf"
 	testDir     = "test"
-	intDir      = "intergration"
+	intDir      = "integration"
 	testFixture = "foo"
 	testExample = "bar"
 	testInvalid = "baz"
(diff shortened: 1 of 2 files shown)


infra/blueprint-test/pkg/discovery/discover_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	baseDir     = "testdata/tf"
+	testDir     = "test"
+	intDir      = "intergration"
```
morgante, line 30: Typo?

```suggestion
	intDir      = "integration"
```

morgante (approved):

infra/blueprint-test/pkg/discovery/discover_test.go
bharathkkb: oops, fixed

PR #1058, opened 2022-02-01
chore: allow running workflow through UI

bharathkkb:
Adds a workflow_dispatch trigger to allow running this workflow via the GH UI
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/update-tooling.yml
+++ b/.github/workflows/update-tooling.yml
@@ -2,6 +2,7 @@ name: Update Tooling
 on:
   schedule:
     - cron: "0 2 * * *"
+  workflow_dispatch:
 
 env:
   TERRAFORM_URL: "https://api.github.com/repos/hashicorp/terraform/releases/latest"


.github/workflows/update-tooling.yml
```diff
@@ -2,6 +2,7 @@ name: Update Tooling
 on:
   schedule:
     - cron: "0 2 * * *"
+  workflow_dispatch:
```
morgante, line 5:
```suggestion
  workflow_dispatch
```

morgante (approved):

.github/workflows/update-tooling.yml
bharathkkb: I believe a colon is required.  Otherwise throws a cannot read implicit mapping error

morgante: It should probably be `workflow_dispatch: {}` then, but if this works it's fine.

PR #1065, opened 2022-02-07
feat: (Blueprint) Allow customization of setup sa key

diegolnasc:
Make setup sa key more flexible for other tests.

+ Minor improvements.

### DIFF
--- a/infra/blueprint-test/pkg/tft/terraform.go
+++ b/infra/blueprint-test/pkg/tft/terraform.go
@@ -162,14 +162,18 @@ func NewTFBlueprintTest(t testing.TB, opts ...tftOption) *TFBlueprintTest {
 	}
 	// load setup sa Key
 	if tft.saKey != "" {
-		tft.loadSetupSAKey(tft.saKey, false)
+		gcloud.ActivateCredsAndEnvVars(tft.t, tft.saKey)
 	}
 	// load TFEnvVars from setup outputs
 	if tft.setupDir != "" {
 		tft.logger.Logf(tft.t, "Loading env vars from setup %s", tft.setupDir)
 		loadTFEnvVar(tft.tfEnvVars, tft.getTFOutputsAsInputs(terraform.OutputAll(tft.t, &terraform.Options{TerraformDir: tft.setupDir, Logger: tft.logger})))
 		if credsEnc, exists := tft.tfEnvVars[fmt.Sprintf("TF_VAR_%s", setupKeyOutputName)]; tft.saKey == "" && exists {
-			tft.loadSetupSAKey(credsEnc, true)
+			if credDec, err := b64.StdEncoding.DecodeString(credsEnc); err == nil {
+				gcloud.ActivateCredsAndEnvVars(tft.t, string(credDec))
+			} else {
+				tft.t.Fatalf("Unable to decode setup sa key: %v", err)
+			}
 		} else {
 			tft.logger.Logf(tft.t, "Skipping credential activation %s output from setup", setupKeyOutputName)
 		}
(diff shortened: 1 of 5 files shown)


infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -60,6 +61,12 @@ func WithName(name string) tftOption {
 	}
 }
 
+func WithSetupSaKey(saEnc string) tftOption {
+	return func(f *TFBlueprintTest) {
+		f.saKey = saEnc
+	}
+}
```
bharathkkb, line 68:
```suggestion
func WithSetupSaKey(saKey string) tftOption {
	return func(f *TFBlueprintTest) {
		f.saKey = saKey
	}
}
```

infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -40,6 +40,7 @@ const setupKeyOutputName = "sa_key"
 type TFBlueprintTest struct {
 	discovery.BlueprintTestConfig                          // additional blueprint test configs
 	name                          string                   // descriptive name for the test
+	saKey                         string                   // optional setup encoded sa key
```
bharathkkb, line 43: If we are specifying an explicit SA key, I would expect it to be already decoded (i.e not b64 encoded) since thats how you would download/generate it via gcloud. WDYT?

infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -157,16 +164,14 @@ func NewTFBlueprintTest(t testing.TB, opts ...tftOption) *TFBlueprintTest {
 	if tft.setupDir != "" {
 		tft.logger.Logf(tft.t, "Loading env vars from setup %s", tft.setupDir)
 		loadTFEnvVar(tft.tfEnvVars, tft.getTFOutputsAsInputs(terraform.OutputAll(tft.t, &terraform.Options{TerraformDir: tft.setupDir, Logger: tft.logger})))
-		// setup credentials if explicit sa_key output from setup
-		credsEnc, exists := tft.tfEnvVars[fmt.Sprintf("TF_VAR_%s", setupKeyOutputName)]
-		if !exists {
-			tft.logger.Logf(tft.t, "Unable to find %s output from setup, skipping credential activation", setupKeyOutputName)
+		if tft.saKey != "" {
+			tft.loadSetupSAKey(tft.saKey)
 		} else {
```
bharathkkb, line 169: This is nested within `if tft.setupDir != ""{..}` so if there is no `setupDir` control wont reach here. We could check and activate SA before this if ` tft.saKey != ""` so it works even without setup.

infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -205,6 +210,15 @@ func (b *TFBlueprintTest) getTFOutputsAsInputs(o map[string]interface{}) map[str
 	return n
 }
 
+// loadSetupSAKey activate a gcp encoded sa key
+func (b *TFBlueprintTest) loadSetupSAKey(saEnc string) {
+	credDec, err := b64.StdEncoding.DecodeString(saEnc)
+	if err != nil {
+		b.t.Fatalf("Unable to decode setup sa key: %v", err)
+	}
```
bharathkkb, line 218: We may need to refactor this based on my comment above if we decide to accept plain SA key for explicit.

infra/blueprint-test/pkg/utils/env.go
```diff
@@ -31,3 +31,11 @@ func ValFromEnv(t testing.TB, k string) string {
 	}
 	return v
 }
+
+// SetEnv set a environment variable.
+func SetEnv(t testing.TB, key string, value string) {
+	err := os.Setenv(key, value)
+	if err != nil {
+		t.Fatal(err)
```
bharathkkb, line 39: Thanks for refactoring! We could also add some context to the error as to which key failed.

bharathkkb (commented):
Thanks for the PR @diegolnasc

infra/blueprint-test/pkg/tft/terraform.go
diegolnasc: I agree.

My script was already generating base64 so I ended up skewing it.

infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -205,6 +210,19 @@ func (b *TFBlueprintTest) getTFOutputsAsInputs(o map[string]interface{}) map[str
 	return n
 }
 
+// loadSetupSAKey activate a gcp encoded sa key
+func (b *TFBlueprintTest) loadSetupSAKey(saKey string, isEncoded bool) {
```
morgante, line 214: Instead of having the isEncoded logic here, I'd just keep it on line 164.

diegolnasc: I thought it would be cleaner in the _func loadSetupSAKey_. 

Well, I'm going to delete the func because it will lose the sense of existing just to call gcloud cmd.

morgante (approved):

PR #1067, opened 2022-02-08
feat: add helper for goldenfiles

bharathkkb:
Adds helper to manage goldenfiles for asserting complex content equality like [GKE](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/35ac28c03141f238c1a4836b150f2c374abd5b94/test/integration/beta_cluster/controls/gcloud.rb#L51-L73) API responses


morgante (approved):
Could also use a lib (ex. https://github.com/bradleyjkemp/cupaloy) but this seems simple enough.

bharathkkb:
@morgante Thanks, I decided to not use as lib as I hope to add some additional functionality to modify certain fields. GCP resource responses sometime have UIDs, createTime etc which we can ignore when saving/asserting with goldenfiles.

Issue #1069, opened 2022-02-10
blueprint test: add gcloud runf


bharathkkb:
Adding a gcloud.Runf() which wraps Run could remove some boilerplate.
```
op = gcloud.Run(t, fmt.Sprintf("sql users list --instance %s", pSql.GetStringOutput("name")))
```
could become
```
op = gcloud.Runf(t, "sql users list --instance %s", pSql.GetStringOutput("name"))
```

bharathkkb:
fixed by https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1070

PR #1070, opened 2022-02-11
feat: add support to runf

diegolnasc:
I'm glad I saw this issue (https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1069). 

I missed this method doing tests.


infra/blueprint-test/pkg/utils/transforms.go
```diff
@@ -40,3 +42,20 @@ func GetResultStrSlice(rs []gjson.Result) []string {
 	}
 	return s
 }
+
+func StringFromTextAndArgs(msgAndArgs ...interface{}) string {
```
bharathkkb, line 46: Since this is only used in gcloud (for now) can we keep it unexported in gcloud? In the future if others need it, we can split it out.

diegolnasc: Done.

PR #1074, opened 2022-02-16
feat: add goldenfile sanitizers

bharathkkb:
This adds sanitization functionality which allows modification of goldenfile data. A few common sanitizers are included in the package for helping us scrub values that change between test runs.

### DIFF
--- a/infra/blueprint-test/pkg/golden/golden.go
+++ b/infra/blueprint-test/pkg/golden/golden.go
@@ -124,6 +124,13 @@ func (g *GoldenFile) ApplySanitizers(s string) string {
 	return s
 }
 
+// ApplySanitizersJSON returns sanitizes and returns JSON result
+func (g *GoldenFile) ApplySanitizersJSON(s gjson.Result) gjson.Result {
+	resultStr := s.String()
+	resultStr = g.ApplySanitizers(resultStr)
+	return utils.ParseJSONResult(g.t, resultStr)
+}
+
 // GetJSON returns goldenfile as parsed json
 func (g *GoldenFile) GetJSON() gjson.Result {
 	return utils.LoadJSON(g.t, g.GetName())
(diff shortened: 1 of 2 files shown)


infra/blueprint-test/pkg/golden/golden.go
```diff
@@ -37,9 +38,26 @@ const (
...
+
+// StringSanitizer replaces all occurrences of old string with new string
+func StringSanitizer(old, new string) Sanitizer {
+	return func(s string) string {
+		old := strings.ReplaceAll(s, old, new)
+		return old
+	}
+}
+
+// ProjectIDSanitizer replaces all occurrences of current gcloud project ID with PROJECT_ID string
```
morgante, line 57: Should we include this by default?

infra/blueprint-test/pkg/golden/golden.go
```diff
@@ -91,6 +118,21 @@ func (g *GoldenFile) GetName() string {
...
+// ApplySanitizers returns sanitized string
+func (g *GoldenFile) ApplySanitizers(s string) string {
+	for _, sanitizer := range g.sanitizers {
+		s = sanitizer(s)
+	}
+	return s
+}
+
+// ApplySanitizersJSON returns sanitizes and returns JSON result
+func (g *GoldenFile) ApplySanitizersJSON(s gjson.Result) gjson.Result {
```
morgante, line 130:
```suggestion
func (g *GoldenFile) GetSanitizedJSON (s gjson.Result) gjson.Result {
```

morgante (changes requested):

infra/blueprint-test/pkg/golden/golden.go
bharathkkb: Done

infra/blueprint-test/pkg/golden/golden.go
bharathkkb: I am leaning towards no as we don't set gcloud config in container today. We can add this as a default later on if it becomes a common pattern.

morgante: What's the behavior if it's not set? (Can we add a test for that?)

Assuming it doesn't cause any errors if unset, I suggest we use it be default.

bharathkkb: @morgante added as default and added a test. If its not set exit code is still 0 with an empty result. I added some additional logging for this case, so users know its skipping.

PR #1077, opened 2022-02-18
chore: Update page to point to the Terraform modules catalog on cloud.google.com

kaariger:


docs/meta/terraform.md.tmpl
```diff
@@ -1,19 +1,5 @@
 <!--- This file is autogenerated, edit the template in meta/ -->
```
morgante, line 1: Instead of modifying this template, we should just stop auto-generating the `terraform.md` file.

morgante (changes requested):

docs/meta/terraform.md.tmpl
kaariger: The generator currently does not auto run but has to be manually triggered. So this file should not be updated unless manually done by running the code.

I can remove that code although I thought leaving the code was ok for now.

bharathkkb: @kaariger how are we generating https://g.co/dev/terraformfoundation? I assumed that was using this template?

kaariger: > @kaariger how are we generating https://g.co/dev/terraformfoundation? I assumed that was using this template?

That is currently being manually maintained.

morgante:
Replaced by #1078.

kaariger:
closing this since its superseded by #1078

docs/meta/terraform.md.tmpl
bharathkkb: Looking at the source of https://g.co/dev/terraformfoundation it does look like the generated txt is slightly different but seems like something we should add to the template.

PR #1083, opened 2022-02-26
feat:adds asmcli to the krm dev tools image

g-awmalik:
This is required for installing asm during int. testing for policy controller.


infra/build/developer-tools-krm/Dockerfile
```diff
@@ -20,6 +20,13 @@ RUN apk update && apk add --no-cache openrc docker-cli docker screen nodejs-curr
 # Additional go tooling
 RUN GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.40.1
 
+# Required to download and install asmcli
+ARG ASMCLI_VERSION
+ENV ASMCLI_VERSION ${ASMCLI_VERSION}
+
+ADD ./build/install_asmcli.sh /build/
+RUN /build/install_asmcli.sh ${ASMCLI_VERSION}
+
```
bharathkkb, line 29: We can remove /build since we wont need it after install

infra/build/Makefile
```diff
@@ -29,6 +29,7 @@ TERRAGRUNT_VERSION := 0.25.3
 KUSTOMIZE_VERSION := 3.6.0
 KPT_VERSION := 1.0.0-beta.13
 CFT_CLI_VERSION := 0.4.2
+ASMCLI_VERSION := 1.12
 
 DOCKER_TAG_VERSION_DEVELOPER_TOOLS := 1.4.1
```
bharathkkb, line 34:
```suggestion
DOCKER_TAG_VERSION_DEVELOPER_TOOLS := 1.4.2
```

bharathkkb (approved):

infra/build/developer-tools-krm/Dockerfile
g-awmalik: is this what you mean?
```
ADD ./build/install_asmcli.sh .
RUN install_asmcli.sh ${ASMCLI_VERSION}
```
Having the install script at root doesn't seem right. Just to be clear this isn't install ASM yet. This is only setting up the utility to install ASM later.

g-awmalik:
@bharathkkb - i updated the version for kubectl as well in the base image.

infra/build/developer-tools-krm/Dockerfile
bharathkkb: No, I meant 

```
ADD ./build/install_asmcli.sh /build/
RUN /build/install_asmcli.sh ${ASMCLI_VERSION}
RUN rm -rf /build
```
This way `/build/install_asmcli.sh` is not persisted as we dont need it after the CLI is installed.

> Having the install script at root doesn't seem right. 

Yeah we can continue to keep the script in /build which we can then clean up.

> Just to be clear this isn't installing ASM yet. This is only setting up the utility to install ASM later.

Yup, understood that this is just installing the cli and not installing ASM.

g-awmalik: Got it. I'll remove that.

g-awmalik:
@bharathkkb - can you take another look? thnx

bharathkkb (approved):

PR #1092, opened 2022-03-08
feat(cli): improve test targeting

bharathkkb:
This improves test targeting by attempting to find the correct test package as opposed to just filtering based on test function names.

For the user this results in
- cleaner log output without `no tests to run` being displayed for test packages that did not contain the test 
- execution time improvement for blueprints (with 9 tests in terraform-google-network 11.547s without targeting reduced to 4.289s with targeting for init stage)


cli/bptest/cmd.go
```diff
@@ -71,18 +71,22 @@ var runCmd = &cobra.Command{
...
-		if err := isValidTestName(flags.testDir, args[0]); err != nil {
-			return err
-		}
 		return nil
 	},
 	RunE: func(cmd *cobra.Command, args []string) error {
-		intTestDir := flags.testDir
+		intTestDir, err := getIntTestDir(flags.testDir)
+		if err != nil {
+			return err
```
morgante, line 79: We should probably return a human-friendly error here.

morgante (approved):
Since this was done for performance reasons, it might be a good idea if we included add Go benchmarks into our tests.

cli/bptest/cmd.go
bharathkkb: done

bharathkkb:
@morgante I opened https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1097 to track adding e2e tests and benchmarks

PR #1110, opened 2022-04-01
feat: add transform func for removing a json value by path

bharathkkb:
This adds a helper for deleting a json value specified by path. This could be useful when used with goldenfiles where we want to assert most values are equal except for one or two values which are non deterministic.

### DIFF
--- a/infra/blueprint-test/pkg/utils/transforms.go
+++ b/infra/blueprint-test/pkg/utils/transforms.go
@@ -19,6 +19,7 @@ package utils
 import (
 	"github.com/mitchellh/go-testing-interface"
 	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 )
 
 // GetFirstMatchResult returns the first matching result with a given k/v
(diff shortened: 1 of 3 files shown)


infra/blueprint-test/pkg/utils/transforms.go
```diff
@@ -40,3 +41,12 @@ func GetResultStrSlice(rs []gjson.Result) []string {
 	}
 	return s
 }
+
+// DeleteFromResult deletes given path from result and returns the modified result
+func DeleteFromResult(t testing.TB, pth string, r gjson.Result) gjson.Result {
+	n, err := sjson.Delete(r.String(), pth)
```
morgante, line 47: I'm surprised this can't be done on `r` directly. Parsing in and out of JSON is inefficient.

morgante (approved):

infra/blueprint-test/pkg/utils/transforms.go
bharathkkb: Unfortunately they decided to split that into sjson which doesn't seem to operate on gjson.Result. We could add optimizations for more than one deletes (if there is a use case in the future) to prevent roundtripping.

PR #1143, opened 2022-05-24
Added all tf validator CI/CD triggers and added additional test variables

melinath:
As part of improving TFV's support for organization and folder level resources, the tests need to know what the folder & org they're in. As long as we're needing to make that kind of change, I also wanted move everything into Terraform (rather than trying to update all of them via clickops.)

### DIFF
--- a/infra/terraform/test-org/ci-triggers/triggers.tf
+++ b/infra/terraform/test-org/ci-triggers/triggers.tf
@@ -53,10 +53,12 @@ resource "google_cloudbuild_trigger" "int_trigger" {
   ignored_files = ["**/*.md", ".gitignore"]
 }
 
-resource "google_cloudbuild_trigger" "tf_validator" {
+resource "google_cloudbuild_trigger" "tf_validator_pull_integration_tests_tf_12" {
+  name = "tf-validator-pull-integration-tests-tf-12"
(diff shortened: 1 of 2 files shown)


infra/terraform/test-org/ci-triggers/triggers.tf
```diff
@@ -67,11 +69,166 @@ resource "google_cloudbuild_trigger" "tf_validator" {
...
+    _TEST_PROJECT      = local.tf_validator_project_id
+    _TEST_FOLDER       = local.tf_validator_folder_id
+    _TEST_ANCESTRY     = local.tf_validator_ancestry
+    _TEST_ORG          = local.org_id
+  }
+
+  filename = ".ci/cloudbuild-tests-integration.yaml"
+}
+
+resource "google_cloudbuild_trigger" "tf_validator_main_integration_tests_tf_13" {
```
bharathkkb, line 128: nit: Looks like this is the same content as above except version- can we for_each though set of `_TERRAFORM_VERSION`. This will also let us make future version edits in one place.

bharathkkb (commented):
minor nit,

infra/terraform/test-org/ci-triggers/triggers.tf
melinath: Made this change in [f90baca](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1143/commits/f90bacac33e3e690cffb306868a852b96705dc00)

bharathkkb (approved):
Applying in a bit

bharathkkb:
This is applied

PR #1145, opened 2022-05-26
Fixed variable definitions

melinath:
- Removed folders/ prefix from folder id var - needs to just be id
- Removed duplicate folders/ prefix in ancestry variable

Basically, org.outputs.folders["key"] is in the format "folders/12345"


bharathkkb (approved):

bharathkkb:
This is applied

PR #1146, opened 2022-05-26
Made ancestry match legacy format

melinath:
tf validator "ancestry path" expects non-plural type names for historical reasons


bharathkkb (approved):

bharathkkb:
this is applied

PR #1155, opened 2022-06-04
fix: retrieve only CFT_CLI releases

apeabody:


.github/workflows/update-tooling.yml
```diff
@@ -33,6 +33,9 @@ jobs:
             elif [ "$tool" == "KPT" ]; then
               # get latest release including pre-releases for kpt, excluding releases of porch
               LATEST_TOOL_VERSION=$(curl -s ${!TOOL_URL} | jq --raw-output '[ .[] | select( .tag_name | contains("porch/") | not )][0].tag_name' | tr -d "v")
+            elif [ "$tool" == "CFT_CLI" ]; then
+              # get latest CFT_CLI release
+              LATEST_TOOL_VERSION=$(curl -s ${!TOOL_URL} | jq --raw-output '[ .[] | select( .tag_name | contains("blueprint-test") | not )][0].tag_name' | tr -d "v")
```
bharathkkb, line 38: nit: Would a better pattern be to select only CLI release?

bharathkkb (approved):

.github/workflows/update-tooling.yml
apeabody: Yes, I was actually thinking about this over the weekend and will switch to "name", as the CLI Release only include the version in their tag_name.

apeabody: @bharathkkb - This now uses "name" with includes "CLI Release", unfortunately I couldn't use that for KPT as its' "name" only includes the version.

PR #1162, opened 2022-06-09
gave folder viewer and security reviewer access to terraform-validator at the projects folder level

melinath:
Folder viewer is necessary because we now recursively fetch the ancestry instead of using project.getAncestry, in order to support non-project resources.

Security reviewer is necessary because CAI conversion requires fetching the remote IAM state to compute the real end state.


infra/terraform/test-org/tf-validator/iam.tf
```diff
@@ -35,8 +35,14 @@ resource "google_project_iam_member" "kokoro_test_1" {
...
 
-resource "google_project_iam_member" "magic_modules_cloudbuild_sa" {
+resource "google_folder_iam_member" "magic_modules_cloudbuild_sa_folder_viewer" {
+  folder = local.projects_folder_id
+  role   = "roles/resourcemanager.folderViewer"
+  member  = local.magic_modules_cloudbuild_sa
+}
+
+resource "google_folder_iam_member" "magic_modules_cloudbuild_sa_security_reviewer" {
   project = module.terraform_validator_test_project.project_id
```
bharathkkb, line 45: Should this be folder?

melinath: yes - fixed in f8d785ce. Thanks!

bharathkkb (approved):
This has been applied

PR #1165, opened 2022-06-10
add support for backend configuration in terraform init

### LINKED ISSUE #1166: blueprint test: Add support for providing backend configuration in terraform `init` command
Allow in the creation of a new `TFBlueprintTest` to provide [partial backend configuration](https://www.terraform.io/language/settings/backends/configuration#partial-configuration) using "command-line key/value pairs" for the terraform `init` command

daniel-cit:
Add support for partial backend configuration in terraform `init`
Fixes #1166


infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -43,6 +43,8 @@ type TFBlueprintTest struct {
 	saKey                         string                   // optional setup sa key
 	tfDir                         string                   // directory containing Terraform configs
 	tfEnvVars                     map[string]string        // variables to pass to Terraform as environment variables prefixed with TF_VAR_
+	backendConfig                 map[string]interface{}   // backend configuration for terraform init
+	migrateState                  bool                     // suppress user confirmation in a migration in terraform init
```
bharathkkb, line 47: Would setting `TF_IN_AUTOMATION` by default help to suppress this?

infra/blueprint-test/examples/simple_tf_module/main.tf
```diff
@@ -22,6 +22,10 @@ provider "null" {
   version = "~> 2.1"
 }
 
+terraform {
+  backend "local" {}
+}
```
bharathkkb, line 27: What happens if a backend is not defined?

infra/blueprint-test/pkg/tft/terraform.go
daniel-cit: Hi @bharathkkb do you mean the flag [TF_IN_AUTOMATION](https://www.terraform.io/cli/config/environment-variables#tf_in_automation) ?

It does not suppress the error saying `If you wish to attempt automatic migration of the state, use "terraform init -migrate-state"`

One alternative would be to remove the `WithMigrateState` function and do it all in a single function always doing the migration

```hcl
func WithBackendConfig(backendConfig map[string]interface{}) tftOption {
	return func(f *TFBlueprintTest) {
		f.backendConfig = backendConfig
		f.migrateState = true
	}
}
```

bharathkkb: Yeah I think that makes sense, we can split it out if a usecase presents itself in the future.

infra/blueprint-test/examples/simple_tf_module/main.tf
bharathkkb: Looks like it just gives a warning if we pass in a partial config to local backend.

infra/blueprint-test/pkg/tft/terraform.go
daniel-cit: `WithMigrateState` merged  into `WithBackendConfig`

bharathkkb (approved):

PR #1182, opened 2022-07-13
chore: Remove tf pytest helper triggers

bharathkkb:
Discussed removing this with @ludoo as have migrated to GHA for tests.


apeabody (approved):
LGTM!

Issue #1190, opened 2022-07-20
blueprint-test/pkg/krmt broken with kpt v1.0.0-beta.16+


apeabody:
kpt [v1.0.0-beta.16](https://github.com/GoogleContainerTools/kpt/releases/tag/v1.0.0-beta.16)+ no longer supports Unchanged status used by blueprint-test/pkg/krmt.

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/f5c2433232d7966085b238416a3a13438b6a0ce0/infra/blueprint-test/pkg/krmt/krm.go#L289

apeabody:
@bharathkkb - WDYT about pinning KPT to [v1.0.0-beta.15](https://github.com/GoogleContainerTools/kpt/releases/tag/v1.0.0-beta.15) till this is resolved?

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/f5c2433232d7966085b238416a3a13438b6a0ce0/infra/build/Makefile#L30

bharathkkb:
@apeabody sounds good to me. Also if this is breaking anything, you should be able to skip DefaultVerify.

apeabody:
> @apeabody sounds good to me. Also if this is breaking anything, you should be able to skip DefaultVerify.

Thanks - Unfortunately all the krmt tests are impacted, so we can't just skip a few.

PR #1194, opened 2022-07-25
fix: pin KPT to v1.0.0-beta.15

apeabody:
* Pin KPT to v1.0.0-beta.15 for #1190


bharathkkb (approved):
Thanks @apeabody. Let's also patch version bump the image.

PR #1198, opened 2022-07-27
add support for Retryable Terraform Errors

### LINKED ISSUE #1195: Custom retry mode for known TF errors
We should introduce a retry option for retrying known flaky failure cases by passing in `RetryableTerraformErrors` map for [tfoptions](https://pkg.go.dev/github.com/gruntwork-io/terratest/modules/terraform#Options)  with some common known retryable errors.

daniel-cit:
Fixes #1195


infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -36,6 +36,13 @@ import (
 
 const setupKeyOutputName = "sa_key"
 
+var (
+	commonRetryableErrors = map[string]string{
```
bharathkkb, line 40: Let's make this public so users can pass it in via option when needed. This will allow us to maintain a common list while other blueprints can extend or pass their own if needed..

infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -36,6 +36,13 @@ import (
...
+		".*FOLDER_TO_DELETE_NON_EMPTY_VIOLATION.*": "Failed to delete non empty folder.",
+		".*SERVICE_DISABLED.*":                     "Required API not enable.",
```
bharathkkb, line 42: Lets add a comment here that API activation is eventually consistent and that even the project_service resource is reconciled there maybe an activation error.
```suggestion
		".*SERVICE_DISABLED.*":                     "Required API not enabled.",
```

infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -169,6 +183,13 @@ func NewTFBlueprintTest(t testing.TB, opts ...tftOption) *TFBlueprintTest {
 			tft.setupDir = setupDir
 		}
 	}
+	// Common retryable errors
+	if tft.retryableTerraformErrors == nil {
+		tft.retryableTerraformErrors = map[string]string{}
+	}
+	for k, v := range commonRetryableErrors {
+		tft.retryableTerraformErrors[k] = v
+	}
```
bharathkkb, line 192: I am leaning towards first adding just the option. We can add this by default later if we see these commonly.

bharathkkb (commented):
Thanks for adding this @daniel-cit

bharathkkb:
@apeabody Could you also add where we saw those? I forgot if it was something common. For not common or repo specific flakiness, I suggest we pass in needed retries via `WithRetryableTerraformErrors` option.

apeabody:
> @apeabody Could you also add where we saw those? I forgot if it was something common. For not common or repo specific flakiness, I suggest we pass in needed retries via `WithRetryableTerraformErrors` option.

We are seeing that when there is a project_id collision when project factory tries to create a new 
"random" project with a project_id which has previously been used.  Most common with terraform-example-foundation, but could happen anywhere where a new project is created as part of the CI.

infra/blueprint-test/pkg/tft/terraform.go
daniel-cit: fixed

infra/blueprint-test/pkg/tft/terraform.go
daniel-cit: removed

infra/blueprint-test/pkg/tft/terraform.go
daniel-cit: Map made public

daniel-cit:
> > @apeabody Could you also add where we saw those? I forgot if it was something common. For not common or repo specific flakiness, I suggest we pass in needed retries via `WithRetryableTerraformErrors` option.
> 
> @bharathkkb We are seeing that when there is a project_id collision when project factory tries to create a new "random" project with a project_id which has previously been used. Most common with terraform-example-foundation, but could happen anywhere where a new project is created as part of the CI.

@apeabody @bharathkkb
This retry will not be able to fix the `googleapi: Error 409: Requested entity already exists, alreadyExists..` error because this is an unrecoverable error.

in the case of projects, if we have a project creation failure, the ID of the project the configuration is trying to create will not change and it will fail again because the retry is done upon the current terraform state, not a clean one.

apeabody (approved):

bharathkkb (approved):

Issue #1208, opened 2022-08-05
terraform-google-kubernetes-engine autogen broken in CFT cft/developer-tools 1.3.10+ 


apeabody:
Starting with CFT v1.3.10 autogen(https://github.com/apeabody/terraform-google-kubernetes-engine/blob/c9361a8206b91983e6121cee36074d0871f27f03/Makefile#L87) isn't working

```
Traceback (most recent call last):
  File "/usr/local/bin/generate_modules.py", line 22, in <module>
    from jinja2 import Environment, FileSystemLoader
  File "/usr/lib/python3.9/site-packages/jinja2/__init__.py", line 12, in <module>
    from .environment import Environment
  File "/usr/lib/python3.9/site-packages/jinja2/environment.py", line 25, in <module>
    from .defaults import BLOCK_END_STRING
  File "/usr/lib/python3.9/site-packages/jinja2/defaults.py", line 3, in <module>
    from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
  File "/usr/lib/python3.9/site-packages/jinja2/filters.py", line 13, in <module>
    from markupsafe import soft_unicode
ImportError: cannot import name 'soft_unicode' from 'markupsafe' (/usr/lib/python3.9/site-packages/markupsafe/__init__.py)
```

bharathkkb:
Nothing stood out to me btw commits and GHA logs have expired. We don't seem to pin jinja https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/build/install_dependencies.sh#L73

apeabody:
Thanks @bharathkkb - I'll dig into it later, but I think it might be related to a mismatch between jinja2 and markupsafe versions: https://markupsafe.palletsprojects.com/en/2.1.x/changes/#version-2-1-0

PR #1212, opened 2022-08-06
fix: update jinja2 and add cookiecutter for deps

apeabody:
* Fixes: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1208
* Add cookiecutter for [terraform-google-module-template](https://github.com/terraform-google-modules/terraform-google-module-template/blob/master/test/test.sh)

### DIFF
--- a/infra/build/developer-tools/build/install_dependencies.sh
+++ b/infra/build/developer-tools/build/install_dependencies.sh
@@ -72,7 +72,7 @@ apk add --no-cache rsync
 # flake8 is used for lint checks
 pip install flake8
 
-# jinja2 are used for lint checks and autogen
+# jinja2 is used for lint checks and autogen
 # Newer that distro version is needed to resolve: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1208
 pip install -U Jinja2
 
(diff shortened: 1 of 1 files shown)


infra/build/developer-tools/build/install_dependencies.sh
```diff
@@ -69,5 +69,12 @@ apk add --no-cache diffutils
...
 apk add --no-cache rsync
 
-# flake8 and jinja2 are used for lint checks
-pip install flake8 jinja2
+# flake8 is used for lint checks
+pip install flake8
+
+# jinja2 are used for lint checks and autogen
+# Newer that distro version is needed to resolve: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1208
+pip install -U Jinja2
```
bharathkkb, line 77: Would it better to pin version to prevent issues like https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1208?

bharathkkb (approved):

infra/build/developer-tools/build/install_dependencies.sh
apeabody: SG, I'll pin to the major version per discussion.

PR #1213, opened 2022-08-06
fix: remove dup Jinja2 pin

apeabody:


apeabody:
@bharathkkb This was the source of https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1208 as it was actually downgrading Jinja2 after the fact, hence the unexpected version mismatch.  

Long term we should probably have have only a single method between: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/build/install_dependencies.sh#L73 and https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/build/data/requirements.txt

bharathkkb (approved):
Thanks for investigating!

PR #1216, opened 2022-08-11
chore(deps): add requests~=2.28

apeabody:


infra/build/developer-tools/build/install_dependencies.sh
```diff
@@ -69,5 +69,9 @@ apk add --no-cache diffutils
 # rsync is needed for check_documentation in task_helper_functions.sh
 apk add --no-cache rsync
 
+# docker is needed for https://github.com/terraform-google-modules/terraform-google-module-template/tests
+apk add --no-cache docker
```
bharathkkb, line 73: Should we just use the krm image which has docker for testing template repo rather than adding to this if we won't use it anywhere else? We can explain rationale in a comment.

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools-krm/Dockerfile

bharathkkb (approved):

infra/build/developer-tools/build/install_dependencies.sh
apeabody: SGTM!

PR #1219, opened 2022-08-12
chore: update module-template org

bharathkkb:
Since module-template is in terraform-google-modules org.


apeabody (approved):
Thanks for catching @bharathkkb! - I just realized that #1209 was behind when merged.

PR #1222, opened 2022-08-16
chore: update branch protection excludes

apeabody:


infra/terraform/test-org/github/branch.tf
```diff
@@ -23,8 +23,9 @@ locals {
     "terraform-google-secured-data-warehouse",
     "terraform-google-secure-cicd",
     "terraform-google-cloud-run",
-    "terraform-google-on-prem",
     "terraform-google-secret-manager",
+    "terraform-google-load-balanced-vms",
+    "terraform-google-three-tier-app",
```
bharathkkb, line 28: nit: it might be good to add a comment to update https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/0e0f9f6437973db6b6930306647a970eeb628870/infra/terraform/test-org/org/locals.tf#L89 until we have support. We could also pull that list via remote state if we don't want it maintained in two places.

bharathkkb (approved):

infra/terraform/test-org/github/branch.tf
apeabody: SG - It's on my list to add a second provider to handle the other org, so I added a comment in locals.tf till then.

PR #1228, opened 2022-08-18
Increase maximum buffer size for cft cli output reader

mauro-cit:
Fixes: #1229 

This PR solves the error we were having when the Buffer.NewScanner was trying to read the JSON file for the tfplan file that will start being used during the Validation with "gcloud beta terraform vet" on the Foundation. 

Without this small change, it was having an error "error reading output: bufio.Scanner: token too long".


cli/bptest/run.go
```diff
@@ -73,6 +73,7 @@ func streamExec(cmd *exec.Cmd) error {
 	go func() {
 		defer wg.Done()
 		scanner := bufio.NewScanner(op)
+		scanner.Buffer(make([]byte, 4096), 10*1024*1024)
```
bharathkkb, line 76: Lets move these to a const and explain rationale for picking these specific sizes.

bharathkkb (changes requested):
Thanks for the PR @mauro-cit

bharathkkb (approved):

PR #1234, opened 2022-08-29
feat: blueprint-test file logger

bharathkkb:
This adds a new logger that can log to disk instead of stdout. This helps keep cloudbuild logs tidier with additional details in disk and uploaded using [CB artifacts](https://cloud.google.com/build/docs/building/store-artifacts-in-cloud-storage#store-artifact-storage).

### DIFF
--- a/infra/blueprint-test/pkg/utils/logger_test.go
+++ b/infra/blueprint-test/pkg/utils/logger_test.go
@@ -1,0 +1,53 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package utils
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewTestFileLogger(t *testing.T) {
+	tests := []struct {
+		name    string
+		content string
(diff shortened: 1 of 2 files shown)


bharathkkb:
cc @daniel-cit

daniel-cit:
@bharathkkb thanks for the PR

one question, is this new logger to be used with

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/4e9b833071f920c6dd3fbb00771522dd4782c543/infra/blueprint-test/pkg/tft/terraform.go#L147

in this case we have code that does not use the custom logger:

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/1b6b73f036bfb3e864141fa14d762354d71b9ab9/infra/blueprint-test/pkg/utils/poll.go#L41

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/1b6b73f036bfb3e864141fa14d762354d71b9ab9/infra/blueprint-test/pkg/utils/poll.go#L43

do they need to be addressed here or in a new PR?

infra/blueprint-test/pkg/utils/logger_test.go
```diff
@@ -0,0 +1,53 @@
...
+			content: "foo",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := assert.New(t)
+			testFile := path.Join(t.TempDir(), fmt.Sprintf("%s.log", tt.name))
+			fl, flClose := NewTestFileLogger(t, testFile)
+			fl.Logf(t, tt.content)
+			flClose(t)
```
daniel-cit, line 44: would actual usage be something like:

```go
fl, flClose := NewTestFileLogger(t, testFile)
bpt := tft.NewTFBlueprintTest(t,
	tft.WithLogger(fl),
)
defer flClose(t)
```

bharathkkb:
> in this case we have code that does not use the custom logger:

This is primarily for TF logging, utils wont be able to use this logger yet. We can re evaluate if this is needed.

infra/blueprint-test/pkg/utils/logger_test.go
bharathkkb: Yup that looks right.

g-awmalik (approved):

PR #1254, opened 2022-10-03
chore: support gcp repos and add renovate.json

apeabody:
* Add branch protection support for gcp repos
* Automatically set branch protection using the repo's default branch
* Add renovate.json to repos

### DIFF
--- a/infra/terraform/modules/branch_protection/main.tf
+++ b/infra/terraform/modules/branch_protection/main.tf
@@ -1,0 +1,64 @@
+/**
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+provider "github" {
+  owner = var.org
+}
+
+locals {
+  repos = setintersection(var.repo_list, toset(data.github_repositories.repos.names))
+}
+
+data "github_team" "admin" {
+  slug = var.admin
+}
+
+data "github_repositories" "repos" {
+  query = "org:${var.org} archived:no"
+}
+
+data "github_repository" "repo" {
+  for_each = local.repos
+  name     = each.value
+}
+
+resource "github_branch_protection" "default" {
+  for_each      = data.github_repository.repo
+  repository_id = each.value.node_id
+  pattern       = each.value.default_branch
+
+  required_pull_request_reviews {
+    required_approving_review_count = 1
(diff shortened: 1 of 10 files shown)


apeabody:
Hi @bharathkkb - PTAL and let me know if you have any preference on the commit_ author or email values.

infra/terraform/test-org/github/files.tf
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+
+resource "github_repository_file" "renovate_json" {
+  for_each            = data.github_repository.repo
+  repository          = each.value.node_id
+  branch              = "master"
+  file                = ".github/renovate.json"
+  commit_message      = "chore: update renovate.json"
+  commit_author       = "CFT Terraform File"
```
bharathkkb, line 23:
```suggestion
  commit_author       = "CFT Bot"
```

infra/terraform/test-org/github/files.tf
```diff
@@ -0,0 +1,63 @@
...
+  commit_email        = "cloud-foundation-bot@google.com"
+  overwrite_on_create = true
+  content             = <<EOT
```
bharathkkb, line 26: It would be better for syntax highlight/IDE support if we store this as a separate json file and read that in this config.

bharathkkb (approved):

infra/terraform/test-org/github/files.tf
apeabody: SG - Done

infra/terraform/test-org/github/files.tf
apeabody: SG - Done

apeabody:
@bharathkkb - PTAL as this was a major re-write to add support for the gcp repos.

infra/terraform/test-org/github/files.tf
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+resource "github_repository_file" "renovate_json_tgm" {
+  for_each            = data.github_repository.repo
+  repository          = each.value.name
+  branch              = each.value.default_branch
+  file                = ".github/renovate.json"
+  commit_message      = "chore: update renovate.json"
+  commit_author       = "CFT Bot"
+  commit_email        = "cloud-foundation-bot@google.com"
```
bharathkkb, line 24: nit: might be good to pull these into a local since we may want to add more config files in the future.

infra/terraform/test-org/github/branch.tf
```diff
@@ -44,7 +53,35 @@ data "github_team" "cft-admins" {
 resource "github_branch_protection" "master" {
   for_each      = data.github_repository.repo
   repository_id = each.value.node_id
-  pattern       = "master"
+  pattern       = each.value.default_branch
+
+  required_pull_request_reviews {
```
bharathkkb, line 58: nit: since we can't do provider for_each, as a follow up it may make sense to pull out a base repo config (labels, branch protection, files etc) as a module that we can then instantiate twice here passing in two explicit providers.

bharathkkb (approved):
Thanks @apeabody!

infra/terraform/test-org/github/branch.tf
apeabody: Curiously labels don't need separate providers, but I abstracted both branch_protection and a generic repo_file as modules.

PR #1255, opened 2022-10-06
feat:adds file path logic for bpmetadata

g-awmalik:
- validates icon file, if present
- generates examples
- generates sub-blueprints


cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+	"strings"
+)
+
+const (
+	regexEx  = ".*/(examples/.*)"
+	regexMod = ".*/(modules/.*)"
+)
+
+func isPathValid(path string) (bool, error) {
+	_, err := os.ReadFile(path)
```
bharathkkb, line 17: Since we don't read any data, we can skip the read and just do a os.Stat.

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+
+func getDirPaths(configPath string, re *regexp.Regexp) ([]BlueprintMiscContent, error) {
+	dirContent, err := os.ReadDir(configPath)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read dir contents: %v", err)
+	}
+
+	var paths []BlueprintMiscContent
+	for _, c := range dirContent {
+		if c.IsDir() {
```
bharathkkb, line 43: We can keep this left aligned by skipping if !c.IsDir() since we only care about dirs.

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+	re := regexp.MustCompile(regexEx)
+	return getDirPaths(configPath, re)
+}
+
+func getModules(configPath string) ([]BlueprintMiscContent, error) {
+	re := regexp.MustCompile(regexMod)
+	return getDirPaths(configPath, re)
+}
+
+func getDirPaths(configPath string, re *regexp.Regexp) ([]BlueprintMiscContent, error) {
```
bharathkkb, line 35: Lets add a comment as to what this function is doing. I assumed it was collecting dirs in `configPath` but seems like it maybe doing more.

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+	var paths []BlueprintMiscContent
+	for _, c := range dirContent {
+		if c.IsDir() {
+			currDirPath := path.Join(configPath, c.Name())
+			// ignore .terraform directories
+			if strings.HasPrefix(c.Name(), ".terraform") {
+				continue
+			}
+
+			targetDirContent, _ := os.ReadDir(currDirPath)
```
bharathkkb, line 50: nit: we are ignoring errors

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+	_, err := os.ReadFile(path)
+	if err != nil {
+		return false, fmt.Errorf("unable to read file at the provided path: %v", err)
+	}
+
+	return true, nil
+}
+
+func getExamples(configPath string) ([]BlueprintMiscContent, error) {
+	re := regexp.MustCompile(regexEx)
```
bharathkkb, line 26: With must compile we can move this to a global
https://pkg.go.dev/regexp#MustCompile

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"fmt"
+	"os"
+	"path"
+	"regexp"
+	"strings"
+)
+
+const (
+	regexEx  = ".*/(examples/.*)"
```
bharathkkb, line 12: nit: for readability
```suggestion
	regexExamples  = ".*/(examples/.*)"
```

cli/bpmetadata/path_test.go
```diff
@@ -0,0 +1,157 @@
...
+		wantErr bool
+	}{
+		{
+			name:  "valid examples",
+			path:  "content/examples",
+			regex: regexEx,
+			want: []BlueprintMiscContent{
+				BlueprintMiscContent{
+					Name:     "terraform",
+					Location: "examples/acm/acm-terraform-blog-part1/terraform",
```
bharathkkb, line 68: Nested examples are tricky. Here the name should probably be `acm-terraform-blog-part1` rather than `terraform`. I am leaning towards just surfacing top level for now (so from [gke](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/examples/acm-terraform-blog-part1) example, this would be `name:acm-terraform-blog-part1, path:examples/acm-terraform-blog-part1`) and figure out if we should just unnest nested examples (I think these should be few)

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+
+			rPaths, _ := getDirPaths(currDirPath, re)
+			if len(rPaths) > 0 {
+				paths = append(paths, rPaths...)
+			}
+		}
+	}
+
+	return paths, nil
```
bharathkkb, line 75: We should stable sort before return

cli/testdata/bpmetadata/content/examples-some-without-tf/examples/acm/acm-terraform-blog-part1/README.md
```diff
@@ -0,0 +1,76 @@
+# Enable ACM features with Terraform - Part 1
```
bharathkkb, line 1: Since this test data is for dir testing purposes, lets remove file contents in all of `contest/*` files and replace with minimal mock content.

cli/bpmetadata/path_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"path"
+	"regexp"
+	"testing"
+)
+
+const (
+	bptestdataPath = "../testdata/bpmetadata"
+)
+
+func TestIcon(t *testing.T) {
```
bharathkkb, line 14:
```suggestion
func TestIsPathValid(t *testing.T) {
```

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+	"regexp"
+	"strings"
+)
+
+const (
+	regexEx  = ".*/(examples/.*)"
+	regexMod = ".*/(modules/.*)"
+)
+
+func isPathValid(path string) (bool, error) {
```
bharathkkb, line 16: fileExists maybe a better name for this

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, fmt.Errorf("unable to read dir contents: %v", err)
+	}
+
+	var paths []BlueprintMiscContent
+	for _, c := range dirContent {
+		if c.IsDir() {
+			currDirPath := path.Join(configPath, c.Name())
+			// ignore .terraform directories
+			if strings.HasPrefix(c.Name(), ".terraform") {
+				continue
```
bharathkkb, line 47: Prefix may not match in the case of nested dir. We could do a contains to make this more robust.

cli/bpmetadata/path.go
g-awmalik: ack

cli/bpmetadata/path.go
g-awmalik: ack

cli/bpmetadata/path.go
g-awmalik: added

cli/bpmetadata/path.go
g-awmalik: ack

cli/bpmetadata/path.go
g-awmalik: done

cli/bpmetadata/path_test.go
g-awmalik: This is a little more nuanced. See [examples](https://github.com/terraform-google-modules/terraform-google-vm/tree/master/examples/compute_instance) for terraform-google-vm. Here we have nested folders with meaning names as opposed to how examples for done for GKE.

IMO, we should keep the nested folder name since this ACM example is an outlier for a myriad of reasons. I've created a [bug](http://b/253292517) to discuss.

cli/bpmetadata/path.go
g-awmalik: ack

cli/testdata/bpmetadata/content/examples-some-without-tf/examples/acm/acm-terraform-blog-part1/README.md
g-awmalik: Condensed as much as I could.

cli/bpmetadata/path_test.go
g-awmalik: ack

cli/bpmetadata/path.go
g-awmalik: ack

cli/bpmetadata/path.go
g-awmalik: It would since `Name()` returns the name of the dir rather than the path. Even if it's nested, the name would start with ".terraform"

cli/bpmetadata/path.go
bharathkkb: > The expression string is a global variable and we can't do MustComplie outside of a func body

What was the issue with using MustCompile outside func as a global var? The `Must*` are meant to be used for global init as mentioned in the [godoc](https://pkg.go.dev/regexp#MustCompile).

cli/bpmetadata/path_test.go
bharathkkb: Good point, let's discuss next week.

cli/bpmetadata/path.go
bharathkkb: Ah yes, I missed that. Thanks for clarifying.

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func getModules(configPath string) ([]BlueprintMiscContent, error) {
+	re := regexp.MustCompile(regexModules)
+	return getDirPaths(configPath, re)
+}
+
+// getDirPaths traverses a given path and looks for directories
+// with TF configs while ignoring the .terraform* directories created and
+// used internally by the Terraform CLI
```
bharathkkb, line 42: nit: We could potentially use filepath.Walk to simplify this with a walkfunc that collects parent of any .tf file in a set.

cli/bpmetadata/path_test.go
```diff
@@ -0,0 +1,158 @@
...
+			re := regexp.MustCompile(tt.regex)
+			got, err := getDirPaths(path.Join(bptestdataPath, tt.path), re)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("getDirPaths() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			for _, wantC := range tt.want {
+				assert.Contains(t, got, wantC)
+			}
```
bharathkkb, line 155: Now that we sort, we can directly use equal to also test order
```suggestion
			assert.Equal(t, tt.want, got)
```

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,101 @@
...
+					Name:     c.Name(),
+					Location: l,
+				}
+
+				paths = append(paths, dirPath)
+				break
+			}
+		}
+
+		rPaths, _ := getDirPaths(currDirPath, re)
```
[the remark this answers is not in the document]
g-awmalik: changed this to use `filepath.Walk` instead

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,101 @@
...
+		if len(rPaths) > 0 {
+			paths = append(paths, rPaths...)
+		}
```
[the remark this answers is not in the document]
g-awmalik: changed this to use `filepath.Walk` instead

cli/bpmetadata/path.go
g-awmalik: done

cli/bpmetadata/path_test.go
g-awmalik: ack

cli/bpmetadata/path_test.go
g-awmalik: Fair enough. Let's discuss and we can change accordingly. b/254079604

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// getDirPaths traverses a given path and looks for directories
+// with TF configs while ignoring the .terraform* directories created and
+// used internally by the Terraform CLI
+func getDirPaths(configPath string, re *regexp.Regexp) ([]BlueprintMiscContent, error) {
+	paths := []BlueprintMiscContent{}
+	err := filepath.Walk(configPath, func(path string, info fs.FileInfo, err error) error {
+		if err != nil {
+			return fmt.Errorf("error accessing examples in the path %q: %v\n", configPath, err)
```
bharathkkb, line 49: nit: newline not necessary
```suggestion
			return fmt.Errorf("error accessing examples in the path %q: %v", configPath, err)
```

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,90 @@
...
+				paths = append(paths, dirPath)
+			}
+			return filepath.SkipDir
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("error accessing examples in the path %q: %v\n", configPath, err)
```
bharathkkb, line 75: nit: newline not necessary
```suggestion
		return nil, fmt.Errorf("error accessing examples in the path %q: %v", configPath, err)
```

cli/bpmetadata/path.go
```diff
@@ -0,0 +1,90 @@
...
+			d := filepath.Dir(path)
+			if l := trimPath(d, re); l != "" {
+				dirPath := BlueprintMiscContent{
+					Name:     filepath.Base(d),
+					Location: l,
+				}
+
+				paths = append(paths, dirPath)
+			}
+			return filepath.SkipDir
```
bharathkkb, line 68: I believe this approach should work for cases when more than one .tf file was discovered but let's add one extra tf file in one of the test cases just to confirm.

bharathkkb (approved):

cli/bpmetadata/path.go
g-awmalik: It does, I tested it with multiple file just didn't add them to keep the diff min. Added one just now to verify.

PR #1256, opened 2022-10-07
Allow terraform-google-conversion to use cft github actions workload identity

melinath:
We'd like to run TFV tests (and some new related tests) from https://github.com/GoogleCloudPlatform/terraform-google-conversion/. We'd also like to switch to github actions for tests for simplicity. In order to have access to the same test environment as we do via cloudbuild, could we potentially use the cft-github-actions workload identity with the tgc repository?

### DIFF
--- a/infra/terraform/test-org/ci-project/sa.tf
+++ b/infra/terraform/test-org/ci-project/sa.tf
@@ -38,5 +38,9 @@ module "oidc" {
       sa_name   = module.service_accounts.service_account.name
       attribute = "attribute.repository/GoogleCloudPlatform/cloud-foundation-toolkit"
     }
+    cft-github-actions-tgc = {
+      sa_name   = module.service_accounts.service_account.name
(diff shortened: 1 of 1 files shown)


infra/terraform/test-org/ci-project/sa.tf
```diff
@@ -38,5 +38,9 @@ module "oidc" {
       sa_name   = module.service_accounts.service_account.name
       attribute = "attribute.repository/GoogleCloudPlatform/cloud-foundation-toolkit"
     }
+    cft-github-actions-tgc = {
+      sa_name   = module.service_accounts.service_account.name
```
bharathkkb, line 42: This SA only has the permissions defined in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/8d731102d9becaaa245ea2587c4ea65b9cdbce40/infra/terraform/test-org/ci-project/sa.tf#L23-L26. It might be better to create a different SA with necessary roles.

PR #1261, opened 2022-10-18
feat:functionality for writing metadata to disk

g-awmalik:


cli/bpmetadata/cmd.go
```diff
@@ -47,27 +51,53 @@ func generate(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("error getting working dir: %w", err)
 	}
 
+	//read existing metadata.yaml, if present
+	bpObj := &BlueprintMetadata{}
+	f, _ := os.ReadFile(metadataFileName)
```
bharathkkb, line 56: We can stat and skip if file not found

cli/bpmetadata/cmd.go
```diff
@@ -47,27 +51,53 @@ func generate(cmd *cobra.Command, args []string) error {
...
+	// if file is present, try to unmarshal it in BlueprintMetadata
+	if f != nil {
+		bpObj, _ = unmarshalMetadata(f)
```
bharathkkb, line 59: If a yaml is found, it should always be valid. Lets throw an error and exit.

cli/bpmetadata/cmd.go
```diff
@@ -47,27 +51,53 @@ func generate(cmd *cobra.Command, args []string) error {
...
+	}
+
+	// if file was present, check if unmarshaling was successful
+	if bpObj == nil && !mdFlags.quiet {
```
bharathkkb, line 63: I think quiet for this usecase is unnecessary since we expect the format to be always valid. If not valid we will exit early rather than do any processing.

cli/bpmetadata/cmd.go
```diff
@@ -170,8 +207,21 @@ func createInfo(bpPath string, readmeContent []byte) (*BlueprintInfo, error) {
...
+func createInterfaces(bpPath string, interfaces *BlueprintInterface) (*BlueprintInterface, error) {
+	i, err := getBlueprintInterfaces(bpPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if interfaces.VariableGroups != nil {
+		i.VariableGroups = interfaces.VariableGroups
+	}
+
```
bharathkkb, line 219: Whats the reasoning for checking incoming interfaces?

cli/bpmetadata/cmd.go
```diff
@@ -203,3 +253,49 @@ func createContent(bpPath string, rootPath string, readmeContent []byte) Bluepri
...
+	if err != nil {
+		return err
+	}
+
+	err = os.WriteFile(metadataFileName, yFile, 0644)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
bharathkkb, line 269:
```suggestion
	return os.WriteFile(metadataFileName, yFile, 0644)
```

cli/bpmetadata/cmd.go
```diff
@@ -203,3 +253,49 @@ func createContent(bpPath string, rootPath string, readmeContent []byte) Bluepri
 
 	return content
 }
+
+func writeMetadata(obj *BlueprintMetadata) error {
```
bharathkkb, line 257: Let's add some test cases so we can visualize the rendered yaml.

cli/bpmetadata/cmd.go
```diff
@@ -203,3 +253,49 @@ func createContent(bpPath string, rootPath string, readmeContent []byte) Bluepri
...
+
+	err = os.WriteFile(metadataFileName, yFile, 0644)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func unmarshalMetadata(f []byte) (*BlueprintMetadata, error) {
```
bharathkkb, line 272: We could include the file read in this fn too for parity with the writeMeta fn which marshals and writes.

cli/bpmetadata/cmd.go
```diff
@@ -203,3 +253,49 @@ func createContent(bpPath string, rootPath string, readmeContent []byte) Bluepri
...
+	var bpObj BlueprintMetadata
+	err := yaml.Unmarshal(f, &bpObj)
```
bharathkkb, line 274: Are comments preserved with yaml? I remember having to kyaml for comments.

cli/bpmetadata/cmd.go
bharathkkb: Additionally we should also validate GVK so we only process correct metadata versions.

cli/bpmetadata/types.go
```diff
@@ -18,27 +18,27 @@ type BlueprintMetadataSpec struct {
...
-	Icon          string
+	ActuationTool BlueprintActuationTool `json:"actuationTool" yaml:"actuationTool"`
+	Description   *BlueprintDescription  `json:",omitempty" yaml:",omitempty"`
+	Icon          string                 `json:",omitempty" yaml:",omitempty"`
 }
 
 // BlueprintContent defines the detail for blueprint related content such as
 // related documentation, diagrams, examples etc.
 type BlueprintContent struct {
 	Diagrams      []BlueprintDiagram
```
bharathkkb, line 32: I suspect we can omitempty quite a few of these.

cli/bpmetadata/cmd.go
g-awmalik: We need to read contents of the file. Even if we stat, we'll have to read the file if its found. This way the code is kept to a min.

cli/bpmetadata/cmd.go
g-awmalik: GVK check is valid. However, I don't think we can assume files are always going to be valid. It is possible that someone may make a syntactical an error while updating labels, variable groups etc.

cli/bpmetadata/cmd.go
g-awmalik: See comment above

cli/bpmetadata/cmd.go
g-awmalik: If the user had added variables groups manually; we should preserve those entries.

cli/bpmetadata/cmd.go
g-awmalik: Will do. b/254323527

cli/bpmetadata/cmd.go
g-awmalik: The file is already read and I wanted to uses that step for both checking status and reading together.

cli/bpmetadata/cmd.go
g-awmalik: They may not be. I'll check

cli/bpmetadata/types.go
g-awmalik: I can take another pass

cli/bpmetadata/cmd.go
g-awmalik: I've modifed `unmarshalMetadata` to include the logic for stat and read. This reads better now.

cli/bpmetadata/cmd.go
g-awmalik: Added a GVK check

cli/bpmetadata/cmd.go
g-awmalik: Added a `force` flag to generate metadata even if the existing file is erroneous.

cli/bpmetadata/cmd.go
g-awmalik: I've added the file read logic within `unmarshalMetadata` now

cli/bpmetadata/cmd.go
g-awmalik: b/2562073 to investigate and implement this.

g-awmalik:
@bharathkkb - can you take another pass at this?

cli/bpmetadata/types.go
```diff
@@ -4,41 +4,43 @@ import "sigs.k8s.io/kustomize/kyaml/yaml"
...
+	Examples      []BlueprintMiscContent `json:",omitempty" yaml:",omitempty"`
 }
 
 // BlueprintInterface the input and output variables for the blueprint
 type BlueprintInterface struct {
-	Variables      []BlueprintVariable
-	VariableGroups []BlueprintVariableGroup
+	Variables []BlueprintVariable
+	// VariableGroups are manually entered
+	VariableGroups []BlueprintVariableGroup `json:"variableGroups" yaml:"variableGroups"`
```
bharathkkb, line 43: This can be optional.

cli/bpmetadata/cmd.go
```diff
@@ -203,3 +239,47 @@ func createContent(bpPath string, rootPath string, readmeContent []byte) Bluepri
...
+	// marshal and write the file
+	yFile, err := yaml.Marshal(obj)
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(metadataFileName, yFile, 0644)
+}
+
+func unmarshalMetadata(bpPath string) (*BlueprintMetadata, error) {
```
bharathkkb, line 253: nit: lets make this and writeMetadata exported functions so other consumers can use them.

bharathkkb (approved):
Looks great!! Some minor comments.

cli/bpmetadata/cmd.go
```diff
@@ -203,3 +239,47 @@ func createContent(bpPath string, rootPath string, readmeContent []byte) Bluepri
...
+	if err != nil {
+		return &bpObj, err
+	}
+
+	currVersion := bpObj.Meta.TypeMeta.APIVersion
+	currKind := bpObj.Meta.TypeMeta.Kind
+
+	//validate GVK for current metadata
+	if currVersion != metadataApiVersion {
+		return &bpObj, fmt.Errorf("found incorrect version for the metadata: %s", currVersion)
```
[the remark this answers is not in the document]
g-awmalik: ack

cli/bpmetadata/cmd.go
```diff
@@ -170,8 +193,21 @@ func createInfo(bpPath string, readmeContent []byte) (*BlueprintInfo, error) {
...
 
-func createContent(bpPath string, rootPath string, readmeContent []byte) BlueprintContent {
-	var content BlueprintContent
+func createInterfaces(bpPath string, interfaces *BlueprintInterface) (*BlueprintInterface, error) {
+	i, err := getBlueprintInterfaces(bpPath)
+	if err != nil {
+		return nil, err
+	}
+
+	if interfaces.VariableGroups != nil {
```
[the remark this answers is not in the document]
g-awmalik: b/255186505

cli/bpmetadata/types.go
```diff
@@ -50,57 +52,59 @@ type BlueprintRequirements struct {
...
 type BlueprintVariable struct {
 	Name        string
-	Description string
-	VarType     string `yaml:"type"`
-	Default     interface{}
-	Required    bool
+	Description string      `json:",omitempty" yaml:",omitempty"`
+	VarType     string      `yaml:"type"`
+	Default     interface{} `json:",omitempty" yaml:",omitempty"`
+	Required    bool        `json:",omitempty" yaml:",omitempty"`
```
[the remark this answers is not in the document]
g-awmalik: yeah removed omitempty from it

cli/bpmetadata/types.go
g-awmalik: ack

cli/bpmetadata/cmd.go
g-awmalik: ack

PR #1266, opened 2022-10-25
add catalog subcmd

bharathkkb:
This adds a catalog subcommand for exploring blueprints. Currently it only lists available blueprints and some basic info. In the future once metadata rolls out, this will be a client consuming that info.

b/254514965


cli/bpcatalog/cmd.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/spf13/viper"
+)
+
+var catalogListFlags struct {
+	format renderFormat
+	sort   sortOption
+}
+
+const (
+	tgmOrg = "terraform-google-modules"
```
g-awmalik, line 17: nit: can we call this tfModulesOrg

cli/bpcatalog/gh.go
```diff
@@ -0,0 +1,110 @@
...
+
+func newGHService(opts ...ghServiceOption) *ghService {
+	ctx := context.Background()
+	ghs := &ghService{
+		client: github.NewClient(nil),
+		ctx:    ctx,
+	}
+	for _, opt := range opts {
+		opt(ghs)
+	}
```
g-awmalik, line 59: Is this doing anything since the body for `type ghServiceOption` is empty?

cli/bpcatalog/tf_test.go
```diff
@@ -0,0 +1,140 @@
...
+				},
+			},
+			want: []string{
+				"terraform-google-foo",
+				"terraform-google-bar",
+			},
+			sortBy: sortCreated,
+		},
+		{
+			name: "simple created",
```
g-awmalik, line 44: all tests have the same name

g-awmalik (approved):
Minor nits.

cli/bpcatalog/tf_test.go
bharathkkb: fixed

cli/bpcatalog/cmd.go
bharathkkb: done

cli/bpcatalog/gh.go
```diff
@@ -0,0 +1,110 @@
+package bpcatalog
```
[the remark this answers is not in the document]
bharathkkb: We could possibly move. Does any thing else need this now? I err on the side of moving to utils unless we need this for some functionality elsewhere.

cli/bpcatalog/gh.go
bharathkkb: Not sure what you meant by empty - it is a type alias for func(*ghService) which can modify a struct of ghService.

PR #1270, opened 2022-10-28
feat: add labels modules and add stale.yml to repo

apeabody:
@bharathkkb - I had to move the `data "github_repositories"` out of the modules to reduce the number of API calls.


infra/terraform/modules/branch_protection/versions.tf
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+terraform {
+  required_providers {
+    github = {
+      source  = "integrations/github"
```
bharathkkb, line 20: nit: Should we pin this a major version incase of any breaking changes?

bharathkkb (approved):

infra/terraform/modules/branch_protection/versions.tf
apeabody: Sure, will do

Issue #1093, opened 2022-03-08
lint test: discover and fail for unused variables 


bharathkkb:
For cases like https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1168 we can catch these at lint time - for example https://github.com/terraform-linters/tflint/blob/master/docs/rules/terraform_unused_declarations.md

morgante:
Could we integrate tflint as a whole? It would be helpful to have easy access to linting rules without having to do it on each repo.

bharathkkb:
Yes, I was thinking of just integrating and maintaining common ruleset in devtools

Issue #1233, opened 2022-08-26
lint test: configurable lint test for samples


bharathkkb:
Samples would like the follow rules to be enforced
- no variables
- no modules

We will likely want to do https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1093 if possible to do this via tflint.

betsy-lichtenberg:
- No backing files (for example [mig.tf](https://github.com/terraform-google-modules/terraform-google-lb-http/blob/master/examples/multi-mig-http-lb/mig.tf) in [this sample](https://github.com/terraform-google-modules/terraform-google-lb-http/tree/master/examples/multi-mig-http-lb)). Each sample should stand alone in main.tf.
- No README files
- No outputs.tf
- No versions.tf

PR #1272, opened 2022-10-28
feat: add tflint to lint toolchain

bharathkkb:
This adds TF lint to our lint tools. Currently there are two sets of rules - one base set for any configs (examples and any other TF configs discovered) and one with more rules for modules. We also skip any TF configs under `test` as there maybe reasons for non conformance. In the future, we can look into more config options like custom rules etc for https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1233

fixes https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1093


infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -307,6 +307,45 @@ function generate_docs() {
     | sort -u)
 }
 
+function check_tflint() {
+  if [[ "${DISABLE_TFLINT:-}" ]]; then
+    echo "DISABLE_TFLINT set. Skipping tflint check."
+    return 0
+  fi
```
bharathkkb, line 314: @apeabody @g-awmalik I was debating btw opt in vs opt out. Considering we would want most modules to onboard to this, I though opt out might be better to accelerate rollout. This does have the drawback that lint could (and likely will) start failing. With opt in, we would need to tackle onboarding repo by repo + clean up flag at some point when all repos have onboarded. Lmk your thoughts.

apeabody: This looks great @bharathkkb! - To my knowledge very few repos are currently tracking CFT `1` or `1.8`, so this shouldn't be immediately effective on the majority repos?  Assuming that is correct, I would make it on by default and we can tackle clean-up at the same time the repo'd CFT version is updated.

apeabody (approved):
Excited to try this out!

PR #1286, opened 2022-11-18
feat: new test strategy for redeploy validation

bharathkkb:
This adds a new `RedeployTest` strategy for ensuring TF blueprints can be deployed n times in a single project.
Each deployment is done in a separate workspace with a deferred cleanup until all the redeployments have been verified.

### DIFF
--- a/infra/blueprint-test/pkg/tft/terraform.go
+++ b/infra/blueprint-test/pkg/tft/terraform.go
@@ -432,3 +432,23 @@ func (b *TFBlueprintTest) Test() {
 	utils.RunStage("apply", func() { b.Apply(a) })
 	utils.RunStage("verify", func() { b.Verify(a) })
 }
+
+// RedeployTest deploys the test n times in separate workspaces before teardown.
+func (b *TFBlueprintTest) RedeployTest(n int) {
+	if b.ShouldSkip() {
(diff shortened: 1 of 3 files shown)


infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -432,3 +432,23 @@ func (b *TFBlueprintTest) Test() {
...
+	if b.ShouldSkip() {
+		b.logger.Logf(b.t, "Skipping test due to config %s", b.Path)
+		b.t.SkipNow()
+		return
+	}
+	a := assert.New(b.t)
+	for i := 1; i <= n; i++ {
+		ws := terraform.WorkspaceSelectOrNew(b.t, b.GetTFOptions(), fmt.Sprintf("test-%d", i))
+		utils.RunStage("init", func() { b.Init(a) })
+		defer func() {
```
g-awmalik, line 447: Why do we need a defer here?

g-awmalik (approved):

infra/blueprint-test/pkg/tft/terraform.go
bharathkkb: It's for deferred cleanup even in case of failures i.e apply failed midway, verify failed etc so resources can still be torn down.

PR #1288, opened 2022-11-18
chore: update tools to GO 1.18

apeabody:
* Update `addlicense` install method for GO 1.18 and version
* Add [CRCMOD](https://cloud.google.com/storage/docs/gsutil/addlhelp/CRC32CandInstallingcrcmod)


infra/build/developer-tools/build/install_addlicense.sh
```diff
@@ -18,5 +18,5 @@ set -u
 
 cd /build
 
-go get -u github.com/google/addlicense@v1.0.0
+go install github.com/google/addlicense@latest
```
bharathkkb, line 21: Can we still pin to a version?

bharathkkb (approved):

infra/build/developer-tools/build/install_addlicense.sh
apeabody: Sure - Pinning it to v1.1.0

PR #1291, opened 2022-11-22
feat: automate codeowners & renovate chores

apeabody:


infra/terraform/modules/codeowners_file/main.tf
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+provider "github" {
+  owner = var.org
+}
```
bharathkkb, line 19: nit: I think provider initialization within modules are deprecated. We can initialize in root and pass to module too
https://developer.hashicorp.com/terraform/language/modules/develop/providers#passing-providers-explicitly

bharathkkb (approved):
Thanks @apeabody!

infra/terraform/modules/codeowners_file/main.tf
apeabody: Sure, moved to the new model.

PR #1301, opened 2022-11-30
fix: updates approach for discovering mod/submods

g-awmalik:
Previously the repo discovery was dependent of root folders being named as `terraform-google-*`. Although that is appropriate, this approach didn't work when that root folder was mounted as a workspace in the dev-tools container. This PR resolves that issue.


infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -305,6 +305,17 @@ function generate_docs() {
   done < <(find_files . -name '*.tf' -print0 \
     | compat_xargs -0 -n1 dirname \
     | sort -u)
+  
+  echo "Generating blueprint metadata"
```
bharathkkb, line 309: I am thinking we should temporarily guard this with an env var so we can control the initial rollout. This is also used within `check_documentation` so linter will fail for any repo tracking dev tools major.

cli/bpmetadata/repo_test.go
```diff
@@ -29,26 +29,12 @@ func TestGetBpRepoPath(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := getBpPathForRepoName(tt.path)
-			if (err != nil) != tt.wantErr {
-				t.Errorf("getBpPathForRepoName() error = %v, wantErr %v", err, tt.wantErr)
-				return
-			}
+			got := getBpRootPath(tt.path)
 			if got != tt.want {
-				t.Errorf("getBpPathForRepoName() = %v, want %v", got, tt.want)
+				t.Errorf("getBpRootPath() = %v, want %v", got, tt.want)
```
bharathkkb, line 37: nit: let's also update the test function name to match.

cli/bpmetadata/repo_test.go
```diff
@@ -29,26 +29,12 @@ func TestGetBpRepoPath(t *testing.T) {
 			want:    "testdata/bpmetadata/terraform-google-bp01",
 			wantErr: false,
 		},
-		{
-			name:    "invalid top level",
```
bharathkkb, line 33: Could we add some test cases with the docker container scenario where IIUC path would be something like `/workspace/modules/foo` to prevent any regressions in future?

cli/bpmetadata/repo.go
```diff
@@ -21,24 +20,20 @@ type repoSource struct {
 
 const (
 	bpUrlRegEx   = "(.*terraform-google-[^/]+)/?(modules/.*)?"
```
bharathkkb, line 22: Do we need this anymore?

infra/build/developer-tools/build/scripts/task_helper_functions.sh
g-awmalik: Done. I also excluded metadata.yaml in the diff until we solve the real problem.

cli/bpmetadata/repo_test.go
g-awmalik: done

cli/bpmetadata/repo_test.go
g-awmalik: added.

cli/bpmetadata/repo.go
g-awmalik: nope, took it out.

bharathkkb (approved):

infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -357,18 +376,18 @@ function check_documentation() {
...
     --exclude '*/.tfvars' \
     /workspace "${tempdir}" >/dev/null 2>/dev/null
-  cd "${tempdir}"
+  cd "${tempdir}/workspace"
   generate_docs >/dev/null 2>/dev/null
+  # TODO: (b/261241276) preserve verion no. for release PR
   diff -r \
     --exclude=".terraform" \
     --exclude=".kitchen" \
-    --exclude=".git" \
```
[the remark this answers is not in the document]
g-awmalik: I took it out to keep the command simple since there won't be any changes to this folder between the time its copied and when the diff is executed.

PR #1327, opened 2022-12-10
chore: add managed conventional-commit-lint.yaml

apeabody:


infra/terraform/test-org/github/resources/conventional-commit-lint.yaml
```diff
@@ -0,0 +1,2 @@
+enabled: true
```
bharathkkb, line 1: nit: headers

bharathkkb (approved):
Thanks @apeabody, should we also add this to https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/modules/branch_protection/main.tf#L40-L46?

apeabody:
> 

Hi @bharathkkb - I was going to soak test first, but I think we got that with: https://github.com/terraform-google-modules/terraform-example-foundation/pull/911 so I'll add to this PR.

apeabody:
> Thanks @apeabody, should we also add this to https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/modules/branch_protection/main.tf#L40-L46?

Hi @bharathkkb - I was going to soak test first, but I think we got that with: https://github.com/terraform-google-modules/terraform-example-foundation/pull/911 so I'll add to this PR now.

PR #1332, opened 2022-12-15
feat: adds tf formatting to module gen helper function

g-awmalik:


bharathkkb (commented):
Thanks @g-awmalik. I am wondering if this would cause an incorrect diff when we do `check_generate_modules` when both format and generate is off since this is running across all modules both autogenerated and manual.

WDYT about a new helper like [fix_lint](https://github.com/GoogleCloudPlatform/blueprints/blob/52230127167ef98f668318637175ab7b691b29b0/utils/lint.sh#L213) like we have in krm which will do fmt, fix_headers, generate etc.

g-awmalik:
I don't think it would. It may cause a diff with existing configs because they may not be formatted but then again they wouldn't pass lint so even that's not valid. If you prefer a helper instead, that's fine too. I'll close this PR in that case.

bharathkkb:
@g-awmalik no preference if you have tested, let's merge this

bharathkkb (approved):

PR #1336, opened 2022-12-16
chore: add lint for infra/terraform

apeabody:


infra/terraform/terraform_backend/terraform.tfstate
```diff
@@ -1,103 +1,96 @@
 {
   "version": 4,
```
bharathkkb, line 2: Not sure why this was checked in  but we could move this our `cft-infra-test-tfstate` backend.

bharathkkb (approved):

infra/terraform/terraform_backend/terraform.tfstate
apeabody: Per discussion removing terraform_backend and retaining the legacy `cloud-foundation-cicd-tfstate` and `cloud-foundation-cicd-seed-projects-tfstate` buckets.

PR #1367, opened 2023-01-12
fix: update blueprint-test for kpt v1.0.0-beta.16+

apeabody:
NOTE: `kpt live apply` no longer includes "unchanged" status, so this somewhat changes the behavior of `DefaultVerify` to verify all are applied successfully.  If the verification of "unchanged" status is used by other consumers (@kaariger) of this package we should revisit the design.

Tested with `go test -v krm_simple_blueprint_test.go`

Fixes: #1190

### DIFF
--- a/infra/blueprint-test/pkg/kpt/kpt.go
+++ b/infra/blueprint-test/pkg/kpt/kpt.go
@@ -61,7 +61,7 @@ func NewCmdConfig(t testing.TB, opts ...cmdOption) *CmdCfg {
 		kOpts.kptBinary = "kpt"
 	}
 	// Validate required KPT version
-	if err := utils.MinSemver(t, "v" + kOpts.RunCmd("version"), MIN_KPT_VERSION); err != nil {
+	if err := utils.MinSemver("v" + kOpts.RunCmd("version"), MIN_KPT_VERSION); err != nil {
 		t.Fatalf("unable to validate minimum required kpt version: %v", err)
 	}
 
--- a/infra/blueprint-test/pkg/kpt/kpt.go
+++ b/infra/blueprint-test/pkg/kpt/kpt.go
@@ -9,7 +9,6 @@ import (
 	"github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/shell"
 	"github.com/mitchellh/go-testing-interface"
-	"golang.org/x/mod/semver"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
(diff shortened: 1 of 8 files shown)


bharathkkb (approved):

infra/blueprint-test/pkg/kpt/kpt.go
```diff
@@ -57,6 +61,17 @@ func NewCmdConfig(t testing.TB, opts ...cmdOption) *CmdCfg {
 		}
 		kOpts.kptBinary = "kpt"
 	}
+	// Validate required KPT version
+	kptVersion, err := utils.KptVersion(kOpts.kptBinary)
```
bharathkkb, line 65: It would be good to keep the kpt specific logic in this package. Couldn't we do `kOpts.RunCmd("version")` to get the version here and maybe the utils could have a general `minSemver(t testing.TB, gotSemver, minSemver string)` func which we call here to assert gotSemver >= minSemver.

apeabody: @bharathkkb - Good idea, I think this is a lot cleaner/reusable.  Let me know if you have anything else before I merge.

infra/blueprint-test/pkg/utils/version.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+
+	"github.com/mitchellh/go-testing-interface"
+	"golang.org/x/mod/semver"
+)
+
+// MinSemver validates gotSemver is not less than minSemver
+func MinSemver(t testing.TB, gotSemver string, minSemver string) error {
```
bharathkkb, line 27: nit: Since we are returning the error back, we can skip passing in the `t`. We usually pass in `t` if we want to error out the test like `t.Error()` within the func.
```suggestion
func MinSemver(gotSemver string, minSemver string) error {
```

infra/blueprint-test/pkg/utils/version.go
```diff
@@ -0,0 +1,38 @@
...
+	if semver.IsValid(gotSemver) != true {
```
bharathkkb, line 28: nit
```suggestion
	if !semver.IsValid(gotSemver) {
```

bharathkkb (approved):

infra/blueprint-test/pkg/utils/version.go
```diff
@@ -0,0 +1,38 @@
...
+		return fmt.Errorf("unable to parse got version %q", gotSemver)
+	} else 	if semver.IsValid(minSemver) != true {
```
apeabody, line 30:
```suggestion
	} else 	if !semver.IsValid(gotSemver) {
```

infra/blueprint-test/pkg/kpt/kpt.go
```diff
@@ -57,6 +60,11 @@ func NewCmdConfig(t testing.TB, opts ...cmdOption) *CmdCfg {
 		}
 		kOpts.kptBinary = "kpt"
 	}
+	// Validate required KPT version
+	if err := utils.MinSemver(t, "v" + kOpts.RunCmd("version"), MIN_KPT_VERSION); err != nil {
```
apeabody, line 64:
```suggestion
	if err := utils.MinSemver("v" + kOpts.RunCmd("version"), MIN_KPT_VERSION); err != nil {
```

PR #1370, opened 2023-01-24
chore: install tinkey in developer-tools image

daniel-cit:
Install [Tinkey ](https://github.com/google/tink/blob/master/docs/TINKEY.md) in developer-tools image

### DIFF
--- a/infra/build/developer-tools/build/install_tinkey.sh
+++ b/infra/build/developer-tools/build/install_tinkey.sh
@@ -20,13 +20,12 @@ cd /build
 
 TINKEY_VERSION=$1
 
-gsutil cp "gs://tinkey/tinkey-${TINKEY_VERSION}.tar.gz" .
-tar -xzvf "tinkey-${TINKEY_VERSION}.tar.gz"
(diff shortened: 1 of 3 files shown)


infra/build/developer-tools/build/install_tinkey.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -e
+set -u
+
+cd /build
+
+TINKEY_VERSION=$1
+
+gsutil cp "gs://tinkey/tinkey-${TINKEY_VERSION}.tar.gz" .
+tar -xzvf "tinkey-${TINKEY_VERSION}.tar.gz"
```
apeabody, line 24: Hi @daniel-cit - I would recommend either downloading, or extracting to a folder.  It will both reduce the rm cleanup steps, and prevent any future upstream additional files from being orphaned.

```suggestion
tar -xzvf "tinkey-${TINKEY_VERSION}.tar.gz" -C ./tinkey
```

apeabody (approved):
Thanks @daniel-cit!

PR #1374, opened 2023-01-30
fix: fixes bp metadata desc issue and adds architecture desc attribute

g-awmalik:


cli/testdata/bpmetadata/md/list-content.md
```diff
@@ -29,3 +29,7 @@ It creates projects and configures aspects like Shared VPC connectivity, IAM acc
 
 ### preDeploy
 To deploy this blueprint you must have an active billing account and billing permissions.
+
+### Architecture
+1- User requests are sent to the front end, which is deployed on two Cloud Run services as containers to support high scalability applications.
+2- The request then lands on the middle tier, which is the API layer that provides access to the backend. This is also deployed on Cloud Run for scalability and ease of deployment in multiple languages. This middleware is a Golang based API.
```
bharathkkb, line 35:
```suggestion
1- User requests are sent to the front end, which is deployed on two Cloud Run services as containers to support high scalability applications.

2- The request then lands on the middle tier, which is the API layer that provides access to the backend. This is also deployed on Cloud Run for scalability and ease of deployment in multiple languages. This middleware is a Golang based API.
```

cli/bpmetadata/types.go
```diff
@@ -62,9 +62,10 @@ type BlueprintActuationTool struct {
 }
 
 type BlueprintDescription struct {
-	Tagline   string `json:",omitempty" yaml:",omitempty"`
-	Detailed  string `json:",omitempty" yaml:",omitempty"`
-	PreDeploy string `json:"preDeploy,omitempty" yaml:"preDeploy,omitempty"`
+	Tagline      string `json:",omitempty" yaml:",omitempty"`
+	Detailed     string `json:",omitempty" yaml:",omitempty"`
+	PreDeploy    string `json:"preDeploy,omitempty" yaml:"preDeploy,omitempty"`
+	Architecture string `json:"architecture,omitempty" yaml:"architecture,omitempty"`
```
bharathkkb, line 68: Should we make this a []string and require an ordered list in md?

cli/testdata/bpmetadata/md/list-content.md
```diff
@@ -29,3 +29,7 @@ It creates projects and configures aspects like Shared VPC connectivity, IAM acc
 
 ### preDeploy
 To deploy this blueprint you must have an active billing account and billing permissions.
+
+### Architecture
```
bharathkkb, line 33: From a UX perspective, embedding the diagram here would let users correlate the diagram and the text.

bharathkkb (approved):
Thanks @g-awmalik some minor comments

cli/testdata/bpmetadata/md/list-content.md
g-awmalik: I plan to change this to an ordered list implementation. Right now the content gets cut off if there's more than one `\n` detected.

cli/bpmetadata/types.go
g-awmalik: Yes, once the implementation is changed, I will change this to an array

cli/testdata/bpmetadata/md/list-content.md
g-awmalik: Yeah I thought about that. Again requires more changes but we should do it. I just wanted this initial fix to go out so we can share a better version of metadata with the partners.

cli/bpmetadata/types.go
g-awmalik: Ended up changing it with this release.

cli/testdata/bpmetadata/md/list-content.md
g-awmalik: Ignore this since now we're looking for an ordered list.

Issue #1353, opened 2023-01-06
Metadata incorrect from forks


bharathkkb:
When users fork and send PRs we still want the metadata to ref the parent repo. https://github.com/terraform-google-modules/terraform-google-sql-db/pull/398/commits/a0689ce520ae3e648f7f8f9fc4ac6d850f78bcc3 we use the fork repo.

PR #1400, opened 2023-02-23
fix: honor repo url when manually authored

g-awmalik:
Fixes: #1353


g-awmalik:
@bharathkkb - can you take a look?

cli/util/git_test.go
```diff
@@ -53,13 +44,46 @@ func TestGetRepoName(t *testing.T) {
...
 			subDir: "modules/bp1",
 			remote: defaultRemote,
-			want:   "bar",
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			dir := tempGitRepoWithRemote(t, tt.repo, tt.remote, tt.subDir)
-			got, err := GetRepoName(dir)
+			_, err := GetRepoUrl(dir)
```
bharathkkb, line 52: Shouldn't we be checking the return from GetRepoUrl to verify it equals tt.repo?

cli/bpmetadata/cmd.go
```diff
@@ -159,30 +159,28 @@ func CreateBlueprintMetadata(bpPath string, bpMetadataObj *BlueprintMetadata) (*
...
 
 	bpMetadataObj.Spec = BlueprintMetadataSpec{
-		Info:         *info,
-		Content:      *content,
-		Interfaces:   *interfaces,
+		Info:         bpMetadataObj.Spec.Info,
+		Content:      bpMetadataObj.Spec.Content,
+		Interfaces:   bpMetadataObj.Spec.Interfaces,
 		Requirements: *requirements,
 	}
```
bharathkkb, line 169: Since the `.create` uses a pointer receiver, bpMetadataObj.Spec.Info/Content/Requirements should already reflect updated content. So IIUC, this could be simplified to
```
bpMetadataObj.Spec.Requirements = *requirements
```

cli/bpmetadata/cmd.go
```diff
@@ -233,23 +231,23 @@ func createInfo(bpPath string, readmeContent []byte) (*BlueprintInfo, error) {
...
 	i, err := getBlueprintInterfaces(bpPath)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	if interfaces.VariableGroups != nil {
-		i.VariableGroups = interfaces.VariableGroups
+	if i.VariableGroups != nil {
+		i.VariableGroups = i.VariableGroups
```
bharathkkb, line 244: I believe this is a noop?

cli/bpmetadata/cmd.go
```diff
@@ -233,23 +231,23 @@ func createInfo(bpPath string, readmeContent []byte) (*BlueprintInfo, error) {
 		i.Icon = iconFilePath
 	}
 
-	return i, nil
+	return nil
 }
 
-func createInterfaces(bpPath string, interfaces *BlueprintInterface) (*BlueprintInterface, error) {
+func (i *BlueprintInterface) create(bpPath string) error {
 	i, err := getBlueprintInterfaces(bpPath)
```
bharathkkb, line 238: Since the pointer receiver is also `i`, wouldn't this shadow it rather than assign it?
I am thinking something like https://go.dev/play/p/iZEdjIAx3Dd may happen.

cli/bpmetadata/cmd.go
g-awmalik: Yes, you're right.

cli/bpmetadata/cmd.go
g-awmalik: Yeah just realized that. We'll have to do something like this: https://go.dev/play/p/N5KOguNzOHz. I've changed it.

cli/bpmetadata/cmd.go
g-awmalik: Yeah, this won't apply since it's already unmarshaled.

cli/bpmetadata/cmd.go
bharathkkb: @g-awmalik In a follow up, could we also add some e2e tests which can catch these?

bharathkkb (approved):

PR #1410, opened 2023-02-23
fix: updates cli tagging for automatic upload to go docs

g-awmalik:


apeabody (approved):

bharathkkb (commented):
Would we need an update in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/3b05220845881efe81ec7fa51c439ac80a589643/.github/workflows/update-tooling.yml#L41-L43?

g-awmalik:
> Would we need an update in
> 
> https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/3b05220845881efe81ec7fa51c439ac80a589643/.github/workflows/update-tooling.yml#L41-L43
> 
> ?

Thanks for catching that @bharathkkb. We would indeed. PTAL now.

PR #1418, opened 2023-03-02
fix: adds policy_id variable to the project factory int trigger

g-awmalik:
Fixes #780

### DIFF
--- a/infra/terraform/test-org/org/locals.tf
+++ b/infra/terraform/test-org/org/locals.tf
@@ -17,6 +17,7 @@
 locals {
   org_id           = "943740911108"
   billing_account  = "01D904-DAF6EC-F34EF7"
+  policy_id        = "203772241124"
   cft_ci_group     = "cft-ci-robots@test.infra.cft.tips"
   gcp_admins_group = "gcp-admins@test.infra.cft.tips"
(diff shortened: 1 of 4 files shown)


infra/terraform/test-org/org/locals.tf
```diff
@@ -17,6 +17,7 @@
 locals {
   org_id           = "943740911108"
   billing_account  = "01D904-DAF6EC-F34EF7"
+  policy_id        = "203772241124"
```
bharathkkb, line 20: This policy ID is not static since some of our tests recreate it. Couldn't we source this during test via gcloud?

g-awmalik: The [`gcloud` invocation](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/build/int.cloudbuild.yaml#L130) is failing for some reason and the `policy_id` is not getting set. 

I think the cloudbuild service agent for the int test build trigger may not have the permissions to read the policy. Can you take a look?

Also, which tests recreate this policy id?

g-awmalik:
This was a permissions issue. Opened [791](https://github.com/terraform-google-modules/terraform-google-project-factory/pull/791).

PR #1420, opened 2023-03-02
fix: adds ctx manager policy reader role to ci bots

g-awmalik:


apeabody (approved):

apeabody (approved):
I assume this has already been applied

g-awmalik:
> I assume this has already been applied

Manually, yes. We can wait for the cron to apply the TF.

apeabody:
> > I assume this has already been applied
> 
> Manually, yes. We can wait for the cron to apply the TF.

SG - I just didn't want to commit as I wasn't sure the status.

PR #1425, opened 2023-03-07
feat: adds UI specific type extensions to bpmetadata

g-awmalik:
- Adds descriptions to existing types.go
- Adds types required for UI rendering
- Reorders existing type definitions to maintain sequence


g-awmalik:
@bharathkkb - PTAL

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
+	CostEstimate BlueprintCostEstimate `json:",omitempty" yaml:",omitempty"`
+	// A list of GCP cloud products used in the solution.
+	// Manually authored.
+	CloudProducts []BlueprintCloudProduct `json:",omitempty" yaml:",omitempty"`
+	// A configuration of fixed and dymanic GCP quotas that apply to the soltuion.
+	// Manually authored.
+	QuotaDetails []BlueprintQuotaDetail `json:",omitempty" yaml:",omitempty"`
+	// Details on the author producing the solution.
+	// Manually authored.
+	Author BlueprintAuthor `json:",omitempty" yaml:",omitempty"`
```
bharathkkb, line 60: nit: Couldn't there be multiple?

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
+	// Autogen details in BlueprintRepoDetail
+	Source *BlueprintRepoDetail
+	// Semantic version  for the packaged solution
+	// Autogenerated: From the `module_name` attribute of the `provider_meta "google"` block
+	// E.g.
+	// provider_meta "google" {
+	//  module_name = "blueprints/terraform/terraform-google-log-analysis/v0.1.5"
+	// }
+	Version string `json:",omitempty" yaml:",omitempty"`
+	// Actuation tool e.g. Terraform and its required version
```
bharathkkb, line 37: nit: since we are planning to use this for godocs let's make sure comments are full sentences ending with a period when applicable for consistency.

cli/bpmetadata/types.go
```diff
@@ -123,6 +172,91 @@ type BlueprintQuotaDetail struct {
...
+
+	// Link to the author's website.
+	Url string `yaml:"url,omitempty"`
+}
+
+type SoftwareGroupType string
+
+const (
+	UnspecifiedSG SoftwareGroupType = "UNSPECIFIED"
+	Os            SoftwareGroupType = "OS"
```
bharathkkb, line 190: nit: acronyms should use all caps if exported, all lower if not exported
```suggestion
	OS            SoftwareGroupType = "OS"
```

cli/bpmetadata/types.go
```diff
@@ -123,6 +172,91 @@ type BlueprintQuotaDetail struct {
...
+	Description string `yaml:"description,omitempty"`
+
+	// Link to the page providing this support option.
+	Url string `yaml:"url"`
+
+	// The organization or group that provides the support option (e.g.:
+	// "Community", "Google").
+	Entity string `yaml:"entity,omitempty"`
+
+	// // Whether to show the customer's support ID.
```
bharathkkb, line 230:
```suggestion
	// Whether to show the customer's support ID.
```

cli/bpmetadata/types.go
```diff
@@ -123,6 +172,91 @@ type BlueprintQuotaDetail struct {
...
+type BlueprintContent struct {
+	// Diagrams are manually entered
+	Architecture  BlueprintArchitecture  `json:"architecture,omitempty" yaml:"architecture,omitempty"`
+	Diagrams      []BlueprintDiagram     `json:",omitempty" yaml:",omitempty"`
+	Documentation []BlueprintListContent `json:",omitempty" yaml:",omitempty"`
+	SubBlueprints []BlueprintMiscContent `json:"subBlueprints,omitempty" yaml:"subBlueprints,omitempty"`
+	Examples      []BlueprintMiscContent `json:",omitempty" yaml:",omitempty"`
+}
+
+// BlueprintInterface the input and output variables for the blueprint
```
bharathkkb, line 245:
```suggestion
// BlueprintInterface defines the input and output variables for the blueprint.
```

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
+package bpmetadata
+
+// BlueprintUI is the top-level structure for holding UI specific metadata
+type BlueprintUI struct {
```
bharathkkb, line 4: Let's add some kind of version info similar to standard

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
...
+	// their sections on the deployment page
+	Input BlueprintUIInput `yaml:"input"`
+
+	// The top-level section for listing runtime (or solution output) information
+	// i.e. the console URL for the VM or a button to ssh into the VM etc based on
+	Runtime BlueprintUIOutput `yaml:"runtime"`
+
+	// The solution version corresponding to the version in the
+	// standard metadata file for keeping data in sync.
+	Version string `yaml:"version"`
```
bharathkkb, line 15: I assume this is just the semver string of the solution? How do we anticipate this going out of sync with standard metadata?

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
...
+	Placeholder string `yaml:"placeholder,omitempty"`
+
+	// Text describing the validation rules for the property. Typically shown
+	// after an invalid input.
+	Validation string `yaml:"validation,omitempty"`
+
+	// The pattern for the value for the input variable
+	Pattern string `yaml:"pattern,omitempty"`
+
+	// Minimum no. of values for the input variable
```
bharathkkb, line 53: Does this only apply for list type var?

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
...
+	Title bool `yaml:"title,omitempty"`
+
+	// A flag to hide or show the property on the UI.
+	Visible bool `yaml:"visible,omitempty"`
+
+	// Property tooltip.
+	Tooltip string `yaml:"tooltip,omitempty"`
+
+	// Placeholder text (when there is no default).
+	Placeholder string `yaml:"placeholder,omitempty"`
```
bharathkkb, line 44: Does this provide additional value over default value in standard?

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Text describing the validation rules for the property. Typically shown
+	// after an invalid input.
+	Validation string `yaml:"validation,omitempty"`
```
bharathkkb, line 48: What is an example value for this? I see some validations like pattern/max/min etc as other fields.

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// The pattern for the value for the input variable
```
bharathkkb, line 50: regex pattern?

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
...
+	// The name of a section to which this property belongs.
+	// Properties belong to the root section if this field is
+	// not set.
+	Section string `yaml:"section,omitempty"`
+
+	// Designates that this property has no impact on the costs, quotas, or
+	// permissions associated with the resources in the expanded deployment.
+	// Typically true for application-specific properties that do not affect the
+	// size or number of instances in the deployment.
+	NoResourceImpact bool `yaml:"noResourceImpact,omitempty"`
```
bharathkkb, line 68: Could we flip this to `ResourceImpact`? I suspect most vars will be ResourceImpact false with only one or two being true so this can reduce verbosity.

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// UI property extension associated with the input variable
+	UIPropertyExtension GooglePropertyExtension `yaml:"x-googleProperty,omitempty"`
+}
+
+// A logical group of properties. [Section][]s may also be grouped into
```
bharathkkb, line 74: > [Section][]s may also be grouped into sub-sections.

DisplaySections seems to be just a slice of []DisplaySection so I was wondering how a subsection might be expressed?

bharathkkb: Ah is it connected via parent?

cli/bpmetadata/types_ui.go
```diff
@@ -0,0 +1,124 @@
...
+	// Accepts string expressions. HTML <code>&lt;a href&gt;</code>
+	// tags only. At most 512 characters.
+	Description string `yaml:"description"`
+
+	// Fixed-width formatted code snippet.
+	// At least one description or snippet is required.
+	// Accepts string expressions. UTF-8 text. At most 512 characters.
+	Snippet string `yaml:"snippet"`
+
+	// If present, this expression determines whether the item is shown.
```
bharathkkb, line 121: What is an example of such an expression?

cli/bpmetadata/types_ui_ext.go
```diff
@@ -0,0 +1,161 @@
...
+
+const (
+	Undefined ExtensionType = "UNDEFINED_TYPE"
+
+	// General formats.
+	EmailAddress    ExtensionType = "EMAIL_ADDRESS"
+	MultiLineString ExtensionType = "MULTI_LINE_STRING"
+
+	// GCE related.
+	GceDiskImage       ExtensionType = "GCE_DISK_IMAGE"
```
bharathkkb, line 13: nit: acronyms capitalized here and throughout
```suggestion
	GCEDiskImage       ExtensionType = "GCE_DISK_IMAGE"
```

cli/bpmetadata/types_ui_ext.go
```diff
@@ -0,0 +1,161 @@
...
+	GceIpForwarding   GceIpForwardingExtension    `yaml:"gceIpForwarding,omitempty"`
+	GceFirewall       GceFirewallExtension        `yaml:"gceFirewall,omitempty"`
+	GceFirewallRange  GceFirewallRangeExtension   `yaml:"gceFirewallRange,omitempty"`
+	GceZone           GceLocationExtension        `yaml:"gceZone,omitempty"`
+	GceRegion         GceLocationExtension        `yaml:"gceRegion,omitempty"`
+	IamServiceAccount IamServiceAccountExtension  `yaml:"iamServiceAccount,omitempty"`
+}
+
+type GceLocationExtension struct {
+	WhitelistedZones   []string `yaml:"whitelistedZones,omitempty"`
```
bharathkkb, line 65: Could we use Allowlisted* instead?

cli/bpmetadata/types_ui_ext.go
```diff
@@ -0,0 +1,161 @@
...
+	NetworkVariable string `yaml:"networkVariable"`
+}
+
+type GceGenericResourceExtension struct {
+	// GCE resource type to be fetched. This field references another
+	// property from the schema, which must have type GceGenericResource.
+	ResourceVariable string `yaml:"resourceVariable"`
+}
+
+type IamServiceAccountExtension struct {
```
bharathkkb, line 157: Are these the roles for the deploying SA (which is already in standard) or is this for something else?

bharathkkb (commented):
Thanks for the PR @g-awmalik

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
-type BlueprintRequirements struct {
-	Roles    []BlueprintRoles
-	Services []string
+	// Title for the solution.
+	// Autogenerated: First H1 text in readme.md
+	Title string
+	// Solution source location and source type
+	// Autogen details in BlueprintRepoDetail
+	Source *BlueprintRepoDetail
+	// Semantic version  for the packaged solution
```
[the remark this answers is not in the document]
g-awmalik: done

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
-
-// BlueprintRequirements defines the roles required and the assocaited services
-// that need to be enabled to provision blueprint resources
-type BlueprintRequirements struct {
-	Roles    []BlueprintRoles
-	Services []string
+	// Title for the solution.
+	// Autogenerated: First H1 text in readme.md
+	Title string
+	// Solution source location and source type
```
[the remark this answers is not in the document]
g-awmalik: I've changed it but we should discuss this.

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
-}
-
-// BlueprintRequirements defines the roles required and the assocaited services
-// that need to be enabled to provision blueprint resources
-type BlueprintRequirements struct {
-	Roles    []BlueprintRoles
-	Services []string
+	// Title for the solution.
+	// Autogenerated: First H1 text in readme.md
+	Title string
```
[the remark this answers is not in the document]
g-awmalik: done

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
+	Detailed string `json:",omitempty" yaml:",omitempty"`
+	// Autogenerated: Markdown after "### PreDeploy"
+	PreDeploy string `json:"preDeploy,omitempty" yaml:"preDeploy,omitempty"`
+	// Autogenerated: Markdown after "### Html"
+	Html string `json:"html,omitempty" yaml:"html,omitempty"`
+	// Autogenerated: Markdown after "### EulaUrls"
+	EulaUrls []string `json:"eulaUrls,omitempty" yaml:"eulaUrls,omitempty"`
+	// Autogenerated: Markdown after "### Architecture"
+	// Deprecated. Use BlueprintContent.Architecture instead.
 	Architecture []string `json:"architecture,omitempty" yaml:"architecture,omitempty"`
```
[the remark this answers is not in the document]
g-awmalik: Removed

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
 }
 
+// A time estimate in secs required for configuring the solution
```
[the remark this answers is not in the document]
g-awmalik: done

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
 type BlueprintTimeEstimate struct {
+	// Autogenerated: Set using the content defined under "### DeploymentTime" E.g.
```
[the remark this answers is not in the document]
g-awmalik: I am tracking this and other TODOs [here](https://taskflow.corp.google.com/workspaces/2562073/work_item/b/267496620?origin=goals&edit=b%2F267496620).

cli/bpmetadata/types.go
```diff
@@ -4,92 +4,141 @@ import (
...
-	LocationKey bool   `json:",omitempty" yaml:",omitempty"`
+	// A top-level (e.g. "Compute Engine") or secondary (e.g. "Binary Authorization")
+	// product used in the solution.
+	ProductId string `json:",omitempty" yaml:",omitempty"`
+	// Url for the product
+	PageUrl string `json:",omitempty" yaml:",omitempty"`
+	// A label string for the product, if it is not an integrated GCP product.
+	// E.g. "Data Studio"
+	Label string `json:",omitempty" yaml:",omitempty"`
+	// If this product is a key in determining the GCP regions/zones the solution
```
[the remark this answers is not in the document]
g-awmalik: It means that this product is limited to certain regions and it should be a "key" in determine the allowed locations for the solution. Although, now that I think about it, I don't think this should be field since we should be able to determine that from the product Id implicitly. I'm going to take this out.

cli/bpmetadata/types.go
g-awmalik: Based on the requirements, it's only one. If there's multiple people involved, that can specified as part of the Author's description.

cli/bpmetadata/types.go
g-awmalik: done

cli/bpmetadata/types.go
g-awmalik: done

cli/bpmetadata/types.go
g-awmalik: done

cli/bpmetadata/types.go
g-awmalik: done

cli/bpmetadata/types_ui.go
g-awmalik: There is a `version` field to track that.

cli/bpmetadata/types_ui.go
g-awmalik: It is the semver of the solution. It's unlikely but could happen in an error scenario where the standard file is updated with the latest variables but the display file is not.

cli/bpmetadata/types_ui.go
g-awmalik: This is a dupe. Validation is enough which will have the validation rules for the value for the variable.

g-awmalik: Going to take this out.

cli/bpmetadata/types_ui.go
g-awmalik: It does because sometimes the placeholder is an instruction and not the value itself.

cli/bpmetadata/types_ui.go
g-awmalik: This is specifically for validating Regex for text fields and ranges for int fields. Min/max is for list type fields and pattern was a dupe.

cli/bpmetadata/types_ui.go
g-awmalik: Yeah but like I mentioned, it was a dupe. Validation covers it.

cli/bpmetadata/types_ui.go
g-awmalik: done

cli/bpmetadata/types_ui.go
g-awmalik: Yes

cli/bpmetadata/types_ui.go
g-awmalik: `outputs().externalIp`.

cli/bpmetadata/types_ui_ext.go
g-awmalik: done

cli/bpmetadata/types_ui_ext.go
g-awmalik: They should be a same but sometimes this extension is used to create SAs with certain roles that are part of the solution. They can have different roles compared to what's required for provisioning.

cli/bpmetadata/types_ui_ext.go
g-awmalik: Yes definitely

bharathkkb (approved):
few minor nits on names/comments

cli/bpmetadata/types_ui.go
g-awmalik: Its more like this: `outputs().hasExternalIP`. I'll add it to the comments.

cli/bpmetadata/types_ui.go
g-awmalik: `BlueprintUI` the top most struct will be part of `BlueprintMetadataSpec` and will inherit the `Kind` and `APIVersion` from it. So there's no need to define it again here. I think I forgot to add `BlueprintUI` to `BlueprintMetadataSpec` in code which may have caused confusion. Will do that now.

cli/bpmetadata/types_ui.go
g-awmalik: Done.

PR #1428, opened 2023-03-08
fix: skips submodule metdata autogen if readme isn't present

g-awmalik:


cli/bpmetadata/cmd.go
```diff
@@ -77,8 +77,22 @@ func generate(cmd *cobra.Command, args []string) error {
...
-	for _, path := range allBpPaths {
-		err = generateMetadataForBpPath(path)
+	for _, modPath := range allBpPaths {
+		// check if module path has readme.md
+		_, err := os.Stat(path.Join(modPath, readmeFileName))
+
+		// throw an error and exit if root module doesn't have a readme.md
+		if !strings.Contains(modPath, modulesPath+"/") && err != nil {
+			return fmt.Errorf("Top-level module does not have a readme. Details: %w\n", err)
+		}
```
bharathkkb, line 87: It seems like if we move this check to after line 59 we can fail fast and potentially simplify the logic?

cli/bpmetadata/cmd.go
```diff
@@ -77,8 +77,22 @@ func generate(cmd *cobra.Command, args []string) error {
...
+
+		// log info if a sub-module doesn't have a readme.md and continue
+		if err != nil {
+			Log.Info("Skiping metadata for sub-module", "Path:", modPath)
```
bharathkkb, line 91: So users are aware of why
```suggestion
			Log.Info("Skipping metadata for sub-module identified as an internal module", "Path:", modPath)
```

bharathkkb (approved):
Thanks @g-awmalik 
Would be good to add a test

cli/bpmetadata/cmd.go
g-awmalik: done

cli/bpmetadata/cmd.go
g-awmalik: done.

PR #1436, opened 2023-03-14
chore: add gtsorbo as network-forensics owner

apeabody:


bharathkkb:
@gtsorbo just wanted to confirm if you were still looking into updating this or if we should hold off?

bharathkkb:
Confirmed with Grant offline

PR #1438, opened 2023-03-14
chore: switch billing

bharathkkb:


bharathkkb:
Not applied yet. After b/273552132 we will try it manually for one or two repos before applying this.

bharathkkb:
This has been applied. I have preserved the old billing account as a few projects are attached and it should not cause issues to already attached projects.

apeabody (approved):

PR #1440, opened 2023-03-17
feat: adds metadata validation logic

g-awmalik:
- Adds a command to generate metadata JSON schema for validation
- Adds a argument to validate metadata(s) in a blueprint
- Enriches metadata types with tags for schema generation


g-awmalik:
@bharathkkb - PTAL. Sorry about the large PR but I couldn't really break this up.

cli/bpmetadata/cmd.go
```diff
@@ -55,6 +58,15 @@ func generate(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("error getting working dir: %w", err)
 	}
 
+	// validate metadata if there is an argument passed into the command
+	if len(args) == 1 {
```
bharathkkb, line 62: Validate seems like it should be a flag? Or maybe we should have verbs as subcommands `generate`/`validate`

cli/bpmetadata/schema/cmd.go
bharathkkb, line 1: Any reason to nest this as a separate package?

cli/bpmetadata/schema/generate.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return err
+	}
+
+	// check if the provided output path is relative
+	if !path.IsAbs(o) {
+		o = path.Join(wdPath, o)
+	}
+
+	Log.Info("generated JSON schema for BlueprintMetadata")
```
bharathkkb, line 34: nit: maybe we the file path would be nice to know in the log?

cli/bpmetadata/schema/generate.go
```diff
@@ -0,0 +1,37 @@
...
+	os.WriteFile(path.Join(o, schemaFileName), sData, 0644)
```
bharathkkb, line 35: nit: let's check error returned from write file

cli/bpmetadata/schema/generate.go
```diff
@@ -0,0 +1,37 @@
+package schema
+
+import (
+	_ "embed"
```
bharathkkb, line 4: It doesn't look like we use the embed directive in this file?

cli/bpmetadata/schema/generate.go
```diff
@@ -0,0 +1,37 @@
...
+
+const schemaFileName = "bpmetadataschema.json"
+
+// generateSchema creates a JSON Schema based on the types
+// defined in the type BlueprintMetadata and it's recursive
+// children. The generated schema will be used to validate
+// all metadata files for consistency and will be uploaded
+// to https://www.schemastore.org/ to provide IntelliSense
+// VSCode for authors manually authoring the metadata.
+func generateSchema(o, wdPath string) error {
```
bharathkkb, line 21: Would be good to add a test for this incase we change `BlueprintMetadata` in such a way that schema can't be generated.

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+	if !path.IsAbs(bpPath) {
+		bpPath = path.Join(wdPath, bpPath)
+	}
+
+	var vErrs []error
+	err := validateMetadataYaml(bpPath, schemaLoader)
+	if err != nil {
+		vErrs = append(vErrs, err)
+		Log.Error("metadata validation failed", "err", err)
+	}
```
bharathkkb, line 36: nit: it would keep the code more left aligned if we do the path collection first i.e `bpPath` + any module path and then in the next block iterate and do `validateMetadataYaml` over all collected paths.

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+
+	modulesPath := path.Join(bpPath, modulesPath)
+	_, err = os.Stat(modulesPath)
+
+	if err != nil {
+		moduleDirs, _ := util.WalkTerraformDirs(modulesPath)
```
bharathkkb, line 42: nit: let's check the error

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+			// log info msg and continue if the file does not exist
+			if err != nil {
+				Log.Info("metadata for module does not exist", "path", d)
+				continue
+			}
+
+			err = validateMetadataYaml(m, schemaLoader)
+			if err != nil {
+				vErrs = append(vErrs, err)
+				Log.Error("metadata validation failed", "err", err)
```
bharathkkb, line 58: Would be good to add path of file in the log

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}
+	}
+
+	if len(vErrs) > 0 {
+		return fmt.Errorf("metadata validation failed for at least one blueprint")
```
bharathkkb, line 64: Could we wrap the vErrs error info too?

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return nil
+}
+
+// validateMetadata validates an individual yaml file present at path "p"
+func validateMetadataYaml(m string, schema gojsonschema.JSONLoader) error {
+
+	// prepare metadata for validation
+	mBytes, _ := ioutil.ReadFile(m)
```
bharathkkb, line 74: nit: error check

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+		return err
+	}
+
+	// we can use a generic interface here since the validation
+	// will be done against the schema built from BlueprintMetadata
+	var obj interface{}
+	if err = json.Unmarshal(mBytes, &obj); err != nil {
+		Log.Error("unable to unmarshal metadata", "error", err)
+		return err
+	}
```
bharathkkb, line 87: What's the reasoning to do this if we don't use `obj` after Unmarshal? Does `gojsonschema.Validate` not throw an error if it is invalid yaml?

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+	yamlLoader := gojsonschema.NewStringLoader(string(mBytes))
+
+	// validate metadata against the schema
+	result, err := gojsonschema.Validate(schema, yamlLoader)
+	if err != nil {
+		return err
+	}
+
+	if !result.Valid() {
+		//var vErrors string
```
bharathkkb, line 99: nit: commented code

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+		return fmt.Errorf("metdata validation failed for file: %s", m)
+	}
+
+	Log.Info("metadata is valid", "path", m)
+	return nil
+}
+
+// prepares metadata bytes for validation since direct
+// validation of YAML is not possible
+func prepareMetadataBytes(b *[]byte) error {
```
bharathkkb, line 114: nit: maybe we should call this yamlToJson?

cli/bpmetadata/validate.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return nil
+}
+
+// validateMetadata validates an individual yaml file present at path "p"
+func validateMetadataYaml(m string, schema gojsonschema.JSONLoader) error {
+
+	// prepare metadata for validation
+	mBytes, _ := ioutil.ReadFile(m)
+	err := prepareMetadataBytes(&mBytes)
```
bharathkkb, line 75: Any particular reason to pass the byte slice ptr and then dereference inside the func? We could change `prepareMetadataBytes` signature to `prepareMetadataBytes(pth string) ([]byte, error)` which would internally read, do the `yaml.YAMLToJSON` and return byte slice.

cli/bpmetadata/validate_test.go
```diff
@@ -0,0 +1,58 @@
...
+		},
+		{
+			name:    "invalid enums for QuotaResourceType",
+			path:    "invalid-metadata-w-enum.yaml",
+			wantErr: true,
+		},
+	}
+
+	// load schema from the binary
+	s := gojsonschema.NewReferenceLoader("file://schema/bpmetadataschema.json")
```
bharathkkb, line 47: Rather than using a static pre generated file, if we generate the metadata as part of the test, that would help us catch any issues when we change the schema.

cli/bpmetadata/cmd.go
g-awmalik: I started off with that but having a subcommand felt excessive. We still have the schema generation as a subcommand but I think want to change that to just a binary on its own since its not something that the enduser will be using. We only need it during the build process to generate the schema. Based on that, I'm OK with making this a flag.

cli/bpmetadata/schema/cmd.go
g-awmalik: Just wanted some isolation but I've changed it to a binary now. I think that works better since generating schema shouldn't be a command.

cli/bpmetadata/schema/generate.go
g-awmalik: done

cli/bpmetadata/schema/generate.go
g-awmalik: done

cli/bpmetadata/schema/generate.go
g-awmalik: Removed

cli/bpmetadata/validate.go
g-awmalik: Done.

cli/bpmetadata/validate.go
g-awmalik: done

cli/bpmetadata/validate.go
g-awmalik: All errs being returned have the file path in them. This should detail the path for the file as well.

cli/bpmetadata/validate.go
g-awmalik: That'll be redundant because whatever is there in vErrs has already been logged as Log.Error a few lines above.

cli/bpmetadata/validate.go
g-awmalik: done

cli/bpmetadata/validate.go
g-awmalik: Removed

cli/bpmetadata/validate.go
g-awmalik: You're right. I was doing some checks on `obj` but removed them and this is just leftover code. `Validate` does everything we need it to do.

cli/bpmetadata/validate.go
g-awmalik: Renamed to `convertYamlToJson`

cli/bpmetadata/validate.go
g-awmalik: I wanted to keep the signature simpler but don't have an objection with what you're suggesting. I've changed it now.

cli/bpmetadata/schema/main.go
```diff
@@ -0,0 +1,44 @@
...
+func main() {
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr,
+			"USAGE: %s [-output=PATH]\n",
+			path.Base(os.Args[0]))
+		flag.PrintDefaults()
+		os.Exit(1)
+	}
+
+	output := flag.String("output", "", "output path for generating the JSON schema definition")
```
bharathkkb, line 24: Since we are not planning to expose this for users, is there value in making the output configurable via flags?

cli/Makefile
```diff
@@ -1,21 +1,25 @@
 SHELL := /bin/bash
 
 # Changing this value will trigger a new release
-VERSION=v0.6.2
+VERSION=v0.7.0
 BINARY=bin/cft
 GITHUB_REPO=github.com/GoogleCloudPlatform/cloud-foundation-toolkit
 PLATFORMS := linux windows darwin
 BUILD_DIR=./bin
+SCHEMA_DIR=./bpmetadata/schema
```
bharathkkb, line 9: Do we use this?

cli/Makefile
```diff
@@ -1,21 +1,25 @@
...
 # Setup the -ldflags option for go build here, interpolate the variable values
 LDFLAGS=-ldflags "-X $(GITHUB_REPO)/cli/cmd.Version=$(VERSION)"
 
+# need an extra build command to embed an most up to date schema in the binary
 .PHONY: build
 build:
 	go build ${LDFLAGS} -o ${BUILD_DIR}/${NAME}
-
+	cd bpmetadata/schema && go build ./...
+	./bpmetadata/schema/schema -output=bpmetadata/schema
```
bharathkkb, line 21: Shouldn't this be before building the CFT binary at L19?

cli/Makefile
```diff
@@ -1,21 +1,25 @@
...
 
 # Setup the -ldflags option for go build here, interpolate the variable values
 LDFLAGS=-ldflags "-X $(GITHUB_REPO)/cli/cmd.Version=$(VERSION)"
 
+# need an extra build command to embed an most up to date schema in the binary
 .PHONY: build
 build:
 	go build ${LDFLAGS} -o ${BUILD_DIR}/${NAME}
-
+	cd bpmetadata/schema && go build ./...
```
bharathkkb, line 20: Rather than building a binary we can just run this as a script
`go run ./bpmetadata/schema/ -output=bpmetadata/schema`

bharathkkb (approved):
Minor nits but LGTM!

cli/bpmetadata/validate_test.go
g-awmalik: Got it. I can do this as a follow-up because I'll need to restructure the packages since call generate right now would create a cyclic dependency.

g-awmalik: #1449

cli/Makefile
g-awmalik: Thanks. Changed it.

cli/Makefile
g-awmalik: We need the metadata built before the schema binary can generate the up to date schema.

cli/Makefile
g-awmalik: We should. I updated the script to use it.

cli/bpmetadata/schema/main.go
g-awmalik: I think there is because we may need to change the location at some point and it's easier if there is already a flag available. There's not much overhead, let's just leave it.

g-awmalik:
Regarding your [comment](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1440#discussion_r1142699098) - that is still a validation failure and _not_ a schema generation failure. The schema will still generate and there's nothing stopping that as long as the code builds.

PR #1450, opened 2023-03-21
fix: adds an addtional build step for schema validation

g-awmalik:


cli/Makefile
```diff
@@ -18,6 +18,7 @@ LDFLAGS=-ldflags "-X $(GITHUB_REPO)/cli/cmd.Version=$(VERSION)"
 build:
 	go build ${LDFLAGS} -o ${BUILD_DIR}/${NAME}
 	go run ./${SCHEMA_DIR} -output=${SCHEMA_DIR}
+	go build ${LDFLAGS} -o ${BUILD_DIR}/${NAME}
```
bharathkkb, line 21: Do we need the go build at L19?

g-awmalik: We don't. Took it out.

bharathkkb (approved):

PR #1451, opened 2023-03-22
feat: adds autogen logic for Blueprint Architecture

g-awmalik:


cli/bpmetadata/markdown_test.go
```diff
@@ -174,3 +174,49 @@ func TestProcessMarkdownContent(t *testing.T) {
...
+		wantErr  bool
+	}{
+		{
+			name:     "Architecture details exists as BlueprintArchitecture",
+			fileName: "list-content.md",
+			title:    "Architecture",
+			want: &BlueprintArchitecture{
+				Description: `1. Step 1
+2. Step 2
+3. Step 3`,
```
bharathkkb, line 193: Looking at the example, would it make more sense as a slice of strings?

bharathkkb (approved):

cli/bpmetadata/markdown_test.go
g-awmalik: Good idea. Changed.

PR #1452, opened 2023-03-23
feat: adds autogen for deployment duration

g-awmalik:


cli/bpmetadata/markdown.go
```diff
@@ -96,3 +112,53 @@ func getMdContent(content []byte, headLevel int, headOrder int, headTitle string
...
+	re := regexp.MustCompile(timeEstimateRegexp)
+	matches := re.FindAllStringSubmatch(durationDetails.literal, -1)
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("unable to find deployment duration")
+	}
+
+	var timeEstimate BlueprintTimeEstimate
+	for _, m := range matches {
+		// each m[2] will have the time in mins
+		i, err := strconv.Atoi(m[2])
```
bharathkkb, line 133: should we also assert len is greater than 2 so we won't go OOB?

cli/bpmetadata/markdown.go
```diff
@@ -96,3 +112,53 @@ func getMdContent(content []byte, headLevel int, headOrder int, headTitle string
...
+
+// getDeploymentDuration creates the deployment and configuration time
+// estimates for the blueprint from README.md
+func getDeploymentDuration(content []byte, headTitle string) (*BlueprintTimeEstimate, error) {
+	durationDetails, err := getMdContent(content, -1, -1, headTitle, true)
+	if err != nil {
+		return nil, err
+	}
+
+	re := regexp.MustCompile(timeEstimateRegexp)
```
bharathkkb, line 124: If we are using `MustCompile`, we can move this to a global var similar to https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/14486905e29e1d4f574a34686a89e54f442636bc/cli/bpmetadata/path.go#L18-L21

bharathkkb (approved):

cli/bpmetadata/markdown.go
g-awmalik: done

PR #1466, opened 2023-03-31
chore: add alloy db, update maintainers and fmt

bharathkkb:


bharathkkb:
this has been applied

infra/terraform/test-org/org/locals.tf
```diff
@@ -108,11 +108,11 @@ locals {
       topics      = join(",", [local.common_topics.compute, local.common_topics.containers])
     },
     {
-      name         = "terraform-ecommerce-microservices-on-gke"
```
apeabody, line 111: We should check why the linter didn't catch originally.

apeabody (approved):
@bharathkkb LGTM!

infra/terraform/test-org/org/locals.tf
bharathkkb: It was my bad, I already had applied the changes in this PR locally so I merged the other `ecommerce-microservices-on-gke` PR and applied without checking lint.

apeabody: No worries @bharathkkb  :) It was just a reminder to myself to investigate as we've had some linter report pass when they actually failed, just happy that wasn't the situtation.

PR #1467, opened 2023-04-03
Add basic boilerplate code for blueprint metadata consumption CLI command

ayushmjain:
Adding boilerplate code for blueprint metadata consumption CLI command.

Usage e.g.: cft blueprint consume jump-start-solutions -p <path-to-metadata.yaml>

This will be used to produce .textproto and .soy files which can be consumed by jump start solutions.

### DIFF
--- a/cli/bpconsume/cmd.go
+++ b/cli/bpconsume/cmd.go
@@ -1,0 +1,22 @@
+package bpconsume
+
+import (
+	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/cli/bpconsume/jumpstartsolutions"
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	Cmd.AddCommand(jumpstartsolutions.Cmd)
+}
+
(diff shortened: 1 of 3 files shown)


prateekj0394 (approved):

g-awmalik:
@ayushmjain @prateekj0394 - do we expect this to become part of the CFT CLI or is this just for demo purposes only? If its just a demo, it may be better to keep things isolated in your forked repo. We can just work off of that. But if this is supposed to live here, I'd like for us to review the solution end-2-end before we can start merging code here. Thoughts?

ayushmjain:
> @ayushmjain @prateekj0394 - do we expect this to become part of the CFT CLI or is this just for demo purposes only? If its just a demo, it may be better to keep things isolated in your forked repo. We can just work off of that. But if this is supposed to live here, I'd like for us to review the solution end-2-end before we can start merging code here. Thoughts?

Sure, we can keep this in our forked repo.

PR #1470, opened 2023-04-03
fix: moves opt-in block for metadata autogen

g-awmalik:
This allows metadata to be generated without setting the `ENABLE_BPMETADATA` env var but at the same time remains opt-in for the `generate_docs` make target.


infra/build/Makefile
```diff
@@ -32,7 +32,7 @@ KUSTOMIZE_VERSION := 3.6.1
 # Updated by Update Tooling Workflow
 KPT_VERSION := 1.0.0-beta.27
 # Updated by Update Tooling Workflow
-CFT_CLI_VERSION := 0.5.2
+CFT_CLI_VERSION := 0.9.0
```
apeabody, line 35: Hmm, looks like we have an issue with `update-tools`: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/actions/runs/4592192569/jobs/8109004037#step:3:73

apeabody (commented):
LGTM!

infra/build/Makefile
apeabody: LoL - You already found it: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1471

g-awmalik: Reverted

g-awmalik:
@apeabody - can you take another look?

infra/build/Makefile
```diff
@@ -49,7 +49,7 @@ OPA_VERSION := 0.42.2
 # Updated by Update Tooling Workflow
 GCRANE_VERSION := 0.13.0
 
-DOCKER_TAG_VERSION_DEVELOPER_TOOLS := 1.10.7
+DOCKER_TAG_VERSION_DEVELOPER_TOOLS := 1.10.8
```
apeabody, line 52: @g-awmalik - I would drop as this currently would be a no-op.  After you release CFT_CLI 0.9.1 in #1475 it should get picked up for inclusion by PR #1472.

apeabody (approved):

infra/build/Makefile
apeabody: OK, I saw that CFT_CLI 0.9.1 was just released, so it's been pulled into #1472. :)

apeabody: Hey @g-awmalik - Sorry for the circle!  I started looking at #1472 and given the bump in TF version we should probably release as `1.11.x`, so you are correct we should bump this to `1.10.8` to ensure it's pulled in by the build process and available for all the repos which are pinned to `1.10`.  I already made the change back.

PR #1471, opened 2023-04-03
fix: updates the update-tooling-workflow for cft and k8s

g-awmalik:
Didn't realize this earlier but we inadvertently introduced a bug as part of 9f87d3c that stopped updating devtools.


apeabody (approved):
Thanks @g-awmalik!

PR #1474, opened 2023-04-04
chore: update install_cft_cli.sh for cli

apeabody:


infra/build/developer-tools/build/install_cft_cli.sh
```diff
@@ -20,5 +20,5 @@ CFT_CLI_VERSION=$1
 
 cd /build
 
-wget "https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/releases/download/v${CFT_CLI_VERSION}/cft-linux-amd64"
+wget "https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/releases/download/cli%2Fv${CFT_CLI_VERSION}/cft-linux-amd64"
```
g-awmalik, line 23: This will fail for v0.5.0 since that is still at v0.5.0 and not cli/v0.5.0. Can you put an if/else to check if `wget` returns a non-zero, `wget` the previous URL else do the new one?

apeabody: Yes, I might need to fiddle with `set -e` as well.

apeabody: OK - It will now first check for the new format, and then fall back to the old if that errors.

apeabody: @g-awmalik - This should be good now.

g-awmalik (approved):

PR #1490, opened 2023-04-14
fix: change codeowners for run and functions

gtsorbo:
cc: @prabhu34


bharathkkb:
this has been applied

PR #1508, opened 2023-04-28
chore: add linter as workflow

apeabody:
Test Example: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/actions/runs/4856014576/jobs/8655130539

This assumes all repos are using the same `./Makefile` format, otherwise falls back to devl-tools `1`.

Removes the current lint as a required check, but otherwise untouched to provide easy roll-back.


apeabody:
@bharathkkb -  given the impact waiting on apply till after initial feedback

infra/terraform/test-org/github/protection.tf
```diff
@@ -135,6 +135,22 @@ module "codeowners_gcp" {
   repos_map = local.gcp_modules_map
 }
 
+module "lint_yaml_tgm" {
```
bharathkkb, line 138: It would be good to have an opt out flag to manage this in repo if any repos additional setup/params

infra/terraform/test-org/github/resources/lint.yaml
```diff
@@ -0,0 +1,39 @@
...
+# limitations under the License.
+
+name: 'lint'
+
+on:
+  pull_request:
+    branches:
+      - 'master'
+      - 'main'
+
```
bharathkkb, line 22: Could we add make this into a tf template and allow passing in env vars? I saw a few repos which pass in custom exclude dirs, disable TF lint etc

https://github.com/terraform-google-modules/terraform-example-foundation/blob/e6089046d93be6a397509ae63b40ff78037a645a/build/lint.cloudbuild.yaml#L21
https://github.com/terraform-google-modules/terraform-google-gcloud/blob/84847e60a758f3d490e84a3efcf20b7a6a73955e/build/lint.cloudbuild.yaml#L20

bharathkkb (approved):
Thanks @apeabody looks great!
Let's give a heads up to the maintainers on Monday before we apply so they are aware

infra/terraform/test-org/github/resources/lint.yaml
apeabody: @bharathkkb - Yes, I thought about using it for the branch using data, but ran out of time last week.

I also noticed at least one repo doesn't put the `Makefile` in `.`: https://github.com/GoogleCloudPlatform/terraform-ecommerce-microservices-on-gke. My current thought is to simply fall back to a sensible default (e.g. `1` or `1.11`), but if needed we could plumb through a version override as well.

infra/terraform/test-org/github/protection.tf
apeabody: add `disable_lint_yaml` to opt out of managed `lint.yaml`

infra/terraform/test-org/github/protection.tf
```diff
@@ -164,7 +180,7 @@ resource "github_branch_protection" "terraform-example-foundation" {
       "cla/google",
       "terraform-example-foundation-int-trigger-default (cloud-foundation-cicd)",
       "terraform-example-foundation-int-trigger-HubAndSpoke (cloud-foundation-cicd)",
-      "terraform-example-foundation-lint-trigger (cloud-foundation-cicd)",
+      "lint / lint (pull_request)",
```
apeabody, line 183: @bharathkkb - I'd rather leave this even if `disable_lint_yaml = true` as this is authoritative and a linter with the same name should be in-place.

infra/terraform/test-org/github/resources/lint.yaml
apeabody: I've mapped through EXCLUDE_LINT_DIRS, set the branch based on the repo's default, and also if no `./Makefile` it will run with `1`.  Do we really want to provide an option to disable tf_lint?

infra/terraform/test-org/github/protection.tf
bharathkkb: Yup sounds good

infra/terraform/test-org/github/resources/lint.yaml
bharathkkb: @apeabody Could we check `./Makefile` and `infra/Makefile`. Between those two I think 99% of cases should be covered.

apeabody: Sure, we could even do something like `$(find . -name Makefile -print -quit)`, the only risk is a repo with multiple `Makefile`, but none in the root, might match the wrong one.

apeabody:
applied - lint / lint checks are n/a

PR #1509, opened 2023-04-29
feat: adds metadata validation to lint tests

g-awmalik:


apeabody (approved):
Looks like this adds `check_metadata()` as disabled, but might still be a good idea to bump the dev-tools patch as the current image will be rebuilt with this added.

g-awmalik:
We'll eventually enable it for all but for now it'll only be enabled for a select few repos.

PR #1518, opened 2023-05-03
chore: gcloud disable remote flags

bharathkkb:


infra/build/developer-tools/build/install_cloud_sdk.sh
```diff
@@ -31,6 +31,7 @@ ln -s /lib /lib64
 gcloud config set core/disable_usage_reporting true
 gcloud config set component_manager/disable_update_check true
 gcloud config set survey/disable_prompts true
+gcloud config set enable_feature_flags false
```
g-awmalik, line 34: Can we put a TODO with an issue so we remember to fix this when the issue resolves?

g-awmalik (approved):

infra/build/developer-tools/build/install_cloud_sdk.sh
bharathkkb: I think this should be permanent since it makes our CI env more hermetic. This is similar to how we have other remote checks disabled like disable_update_check and disable_usage_reporting.

PR #1523, opened 2023-05-04
chore: add go lint

apeabody:


apeabody:
@bharathkkb After the initial commit this will only run if files within the applicable directory are updated: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1523/files#diff-1d3c602e0f80db2572bd6c2b171aae71dca909294527da8471438224cb29675dR36

apeabody:
The new `lint / lint` check results are expected.

apeabody:
> Thanks @apeabody! Could we also add a Makefile target for both projects that run `golangci-lint` locally so we can iterate locally and also fix existing errors.
> 
> ```
> docker run --rm -v $(pwd):/cli -w /cli  golangci/golangci-lint:v1.52.2 golangci-lint run -v
> ```

Sure thing - added `make docker_go_lint` for each.

PR #1531, opened 2023-05-06
chore: support any linter ENVs

apeabody:
* Any lint ENVs can now be added under `lint_env `, for example `EXCLUDE_LINT_DIRS` or `DISABLE_TFLINT`.
* Fix Makefile check with the change from static location to find result.


infra/terraform/modules/lint_file/lint.yaml.tftpl
```diff
@@ -32,14 +35,16 @@ jobs:
...
             echo dev-tools=gcr.io/cloud-foundation-cicd/cft/developer-tools:1 >> "$GITHUB_OUTPUT"
           else
             VERSION=$(grep "DOCKER_TAG_VERSION_DEVELOPER_TOOLS := " $MAKEFILE | cut -d\  -f3)
             IMAGE=$(grep "DOCKER_IMAGE_DEVELOPER_TOOLS := " $MAKEFILE | cut -d\  -f3)
             REGISTRY=$(grep "REGISTRY_URL := " $MAKEFILE | cut -d\  -f3)
             echo dev-tools=$${REGISTRY}/$${IMAGE}:$${VERSION} >> "$GITHUB_OUTPUT"
           fi
-      - run: docker run --rm -e EXCLUDE_LINT_DIRS -v $${{ github.workspace }}:/workspace $${{ steps.variables.outputs.dev-tools }} /usr/local/bin/test_lint.sh%{if exclude_lint_dirs != null }
+      - run: docker run --rm %{ for key, value in lint_env ~}-e ${key} %{ endfor ~}-v $${{ github.workspace }}:/workspace $${{ steps.variables.outputs.dev-tools }} /usr/local/bin/test_lint.sh%{if lint_env != null }
         env:
```
bharathkkb, line 47: If there are no k/vs in lint_env, is an empty `env:` still valid?

bharathkkb (approved):

infra/terraform/modules/lint_file/lint.yaml.tftpl
apeabody: Hmm, I'm not sure how that is handled in `tftpl`, however if needed we can wrap in `%{if lint_env != null } %{ endif }`.

I used `if/endif` in the section below to drop the entire `env:` block.

apeabody: Same handling style as normal tf, so we'll check != null.

apeabody:
applied

PR #1534, opened 2023-05-08
fix: adds unspecified enum value for IP type for bpmetadata

g-awmalik:


g-awmalik:
Will address linter issues as part of [this](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1535) in a  future PR.

bharathkkb:
@g-awmalik it might also be good to add a proto linter so we catch some of these like no unspecified

PR #1541, opened 2023-05-10
fix: bump GO modules and address lint

### LINKED ISSUE #1536: Address linter issues for infra/blueprint-test

apeabody:
Fixes #1536

Tested locally with `go test`


apeabody:
@bharathkkb - I think this was a bug, but let me know if I misunderstood the intention: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1541/files#diff-c6acc5475eea7a39fa709e7981e723f9ddd138438088b2e044cd484c9b013303L94

infra/blueprint-test/pkg/discovery/config.go
```diff
@@ -49,11 +49,11 @@ func GetTestConfig(path string) (BlueprintTestConfig, error) {
 
 // isValidTestConfig validates a given BlueprintTestConfig
 func isValidTestConfig(b BlueprintTestConfig) error {
-	if b.APIVersion != blueprintTestAPIVersion {
-		return fmt.Errorf("invalid APIVersion %s expected %s", b.APIVersion, blueprintTestAPIVersion)
+	if b.ResourceMeta.APIVersion != blueprintTestAPIVersion {
```
bharathkkb, line 52: Did the linter throw an error for these? `BlueprintTestConfig` embeds `yaml.ResourceMeta` so ` b.APIVersion` should be correct AFAIK since it becomes a promoted field.

bharathkkb (approved):
Thanks for patching this up @apeabody 

> I think this was a bug, but let me know if I misunderstood the intention

Yeah that is def a bug

infra/blueprint-test/pkg/discovery/config.go
apeabody: Yup, valid `go`, however changed for the linter which was raising `typecheck: type <struct> has no field or method <Field on embedded struct from another package>` https://github.com/golangci/golangci-lint/issues/826

PR #1550, opened 2023-05-12
chore: remove unused scripts

bharathkkb:
- docs/ is unused as we have switched generation to cli
- infra/concourse is not being used
- infra/utils/delete-projects is not being used - we use [project cleaner](https://github.com/terraform-google-modules/terraform-google-scheduled-function/tree/master/modules/project_cleanup)


apeabody (approved):
LGTM!

PR #1566, opened 2023-05-13
chore: remove dead links from #1550

JanCVanB:
`docs/` is gone, so this removes/redirects references to it


apeabody (approved):
Thanks @JanCVanB!

PR #1568, opened 2023-05-15
chore: add blueprint-contributors team

apeabody:


infra/terraform/modules/team/main.tf
```diff
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+locals {
+  teams = distinct(flatten([for value in var.repos_map : value.groups if try(value.groups, null) != null]))
+  users = distinct(flatten([for value in var.repos_map : value.owners if try(value.groups, null) != null]))
```
bharathkkb, line 19: Shouldn't this be owners?
```suggestion
  users = distinct(flatten([for value in var.repos_map : value.owners if try(value.owners, null) != null]))
```

bharathkkb (approved):

infra/terraform/modules/team/main.tf
apeabody: Great Catch!

apeabody:
applied

PR #1572, opened 2023-05-15
feat: go lint issues and metadata schema updates

g-awmalik:
- addresses go linter issues across the CLI
- adds a field to designate a blueprint as deployable single or multiple times in a project
- changes the min/max fields from int to float


cli/bpmetadata/markdown_test.go
```diff
@@ -212,7 +212,7 @@ func TestProcessArchitectureContent(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			content, err := os.ReadFile(path.Join(mdTestdataPath, tt.fileName))
```
apeabody, line 215: nit: Might want to also check the `os.ReadFile`s for errors.  If could provide a more useful error message than depending on the subsequent `getArchitctureInfo`.

cli/bpmetadata/types_ui.go
```diff
@@ -58,10 +58,10 @@ type DisplayVariable struct {
 	MaximumLength int `json:"maxLength,omitempty" yaml:"maxLength,omitempty"`
 
 	// Minimum value for numeric types.
-	Minimum int `json:"min,omitempty" yaml:"min,omitempty"`
+	Minimum float32 `json:"min,omitempty" yaml:"min,omitempty"`
```
apeabody, line 61: I'm assuming the change to `min` and `max` are to resolve an existing type match issue.

apeabody (approved):

cli/bpmetadata/markdown_test.go
g-awmalik: ack

cli/bpmetadata/types_ui.go
g-awmalik: Yeah although, the type isn't being used yet...

PR #1582, opened 2023-05-17
chore: enable periodics on three-tier

bharathkkb:
Enabled enable periodics for another repo for testing (cc @tpryan ). Will add reporter GHA workflow out of band for now until we do some testing.


bharathkkb:
This has been applied

PR #1594, opened 2023-05-19
chore: use local linter for terraform-google-module-template

apeabody:


apeabody:
Instead of deleting the linter we should remove from tf state

infra/terraform/test-org/org/locals.tf
```diff
@@ -391,9 +391,10 @@ locals {
       topics      = local.common_topics.db
     },
     {
-      name        = "terraform-google-module-template"
-      org         = "terraform-google-modules"
-      description = "Provides a template for creating a Cloud Foundation Toolkit Terraform module"
+      name              = "terraform-google-module-template"
+      org               = "terraform-google-modules"
+      description       = "Provides a template for creating a Cloud Foundation Toolkit Terraform module"
+      disable_lint_yaml = false
```
g-awmalik, line 397: Isn't this going to add the lint yaml to the repo when it's already there?

apeabody: Yup, wrong `bool` value :)

apeabody:
applied with manual state adjustment

Issue #1545, opened 2023-05-11
Support optional periodic tests


bharathkkb:
Allow authors to configure optional period tests for their solution repos. For lint we can likely inject a cron in GHA. For int we could orchestrate trigger exec via scheduler. We might need to brainstorm how we want failures to be surfaced back, maybe GH issues?

apeabody:
> For lint we can likely inject a cron in GHA.

@bharathkkb - Do we want to do automatically for every repo, or expose as a bool in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/test-org/org/locals.tf#L57?

Something like `actions/github-script` can be used to open a issue for the linter.

bharathkkb:
@apeabody I think we should add it as an optional bool but disabled by default. 

> Something like actions/github-script can be used to open a issue for the linter.

 I was thinking of a unified way for both lint and int failures. Maybe we can have another cron that runs from main ~1h after nightly cron and gets the github check status using  `actions/github-script`.

PR #1596, opened 2023-05-19
chore: add periodic reporter workflow

bharathkkb:
fixes https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1545


bharathkkb:
This has been applied

infra/terraform/test-org/github/protection.tf
```diff
@@ -142,14 +142,14 @@ module "codeowners_gcp" {
 }
 
 module "lint_yaml_tgm" {
```
apeabody, line 144: Would be ideal to rename these and move the state: `terraform state mv module.lint_yaml_tgm module.workflow_files`

apeabody (approved):

apeabody (approved):

infra/terraform/test-org/github/protection.tf
bharathkkb: Yeah we could use moved blocks. I will test it out in a follow up since it is non blocking.

PR #1599, opened 2023-05-22
chore: adding enable metadata env var for lint testing

g-awmalik:


bharathkkb (approved):
let's apply this before merge and validate

apeabody:
> @apeabody - this is not setting the envar for the lint test. I applied and tried testing on [this PR](https://github.com/GoogleCloudPlatform/terraform-large-data-sharing-java-webapp/pull/21) but it didn't. Further examination of `lint.yaml` reveals that this map isn't getting passed to the container either. Am I missing something here?

Hey @g-awmalik - In addition to apply on `org`, did you then apply `github`?  The actual lint workflow files will need to be updated for these repos.

g-awmalik:
> Hey @g-awmalik - In addition to apply on `org`, did you then apply `github`? The actual lint workflow files will need to be updated for these repos.

Thanks @apeabody that solves one problem. However, with JSS repos we have the TF configs (and other related assets) under the infra/ folder but the docker mount is at the root. That is all also inconsistent with where the Makefile is present locally i.e. under the infra/ folder. Here's an [example](https://github.com/GoogleCloudPlatform/terraform-large-data-sharing-java-webapp/tree/main/infra).

Can we make this mount configurable? I.e. for JSS we add an attribute to the repo object. If present we mount it to that folder otherwise keep it at root?

apeabody:
> > Hey @g-awmalik - In addition to apply on `org`, did you then apply `github`? The actual lint workflow files will need to be updated for these repos.
> 
> Thanks @apeabody that solves one problem. However, with JSS repos we have the TF configs (and other related assets) under the infra/ folder but the docker mount is at the root. That is all also inconsistent with where the Makefile is present locally i.e. under the infra/ folder. Here's an [example](https://github.com/GoogleCloudPlatform/terraform-large-data-sharing-java-webapp/tree/main/infra).
> 
> Can we make this mount configurable? I.e. for JSS we add an attribute to the repo object. If present we mount it to that folder otherwise keep it at root?

Hi @g-awmalik - The linter does a ["find"](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/modules/workflow_files/lint.yaml.tftpl#L37) for the Makefile, so being under `/infra/` is supported.  We could add the capability to pass a path, but that would also skip linting (headers, training space, etc) the items under .github, the root folder, etc.  Is there a particular issue with the linter? Most of the checks should be path agnostic, but perhaps we can path a path more strategically as an ENV variable to the check in question.

g-awmalik:
> Hi @g-awmalik - The linter does a ["find"](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/modules/workflow_files/lint.yaml.tftpl#L37) for the Makefile, so being under `/infra/` is supported. We could add the capability to pass a path, but that would also skip linting (headers, training space, etc) the items under .github, the root folder, etc. Is there a particular issue with the linter? Most of the checks should be path agnostic, but perhaps we can path a path more strategically as an ENV variable to the check in question.
> 
> We can also exclude a particular directory from the linter if needed using `EXCLUDE_LINT_DIRS`.

The issue with the metadata validation check which starts checking for metadata file at the current path or the path passed to it. Currently, there is no path passed and the it attempts to validate metadata at the root and it doesn't find anything and exits whereas the metadata files are under infra/. We could pass a path to `test_lint.sh` but that will require modify bash scripts and then additional logic for when to pass that path (for JSS) and when to not.

You make a good point regarding linting the root folder in addition to infra/ and I understand the value of that. Maybe we should change the way the metadata validation works i.e. let it walk entire hierarchy and validate metadata file whenever it finds them. Let me look into that and get back.

apeabody:
> > Hi @g-awmalik - The linter does a ["find"](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/modules/workflow_files/lint.yaml.tftpl#L37) for the Makefile, so being under `/infra/` is supported. We could add the capability to pass a path, but that would also skip linting (headers, training space, etc) the items under .github, the root folder, etc. Is there a particular issue with the linter? Most of the checks should be path agnostic, but perhaps we can path a path more strategically as an ENV variable to the check in question.
> > We can also exclude a particular directory from the linter if needed using `EXCLUDE_LINT_DIRS`.
> 
> The issue is with the metadata validation check which starts checking for metadata file(s) at the current path or the path passed to it. Currently, there is no path passed and it attempts to validate metadata at the root, it doesn't find anything and exits whereas the metadata files are under infra/. We could pass a path to `test_lint.sh` as infra/ but that will require modifying bash scripts and then additional logic for when to pass that path (for JSS) and when to not.
> 
> You make a good point regarding linting the root folder in addition to infra/ and I understand the value of that. Maybe we should change the way the metadata validation works i.e. let it walk entire hierarchy and validate metadata file(s) whenever it finds them. Let me look into that and get back.

Got it - Yeah, might be able to do something around here: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/cli/bpmetadata/cmd.go#LL82C1-L82C1

g-awmalik:
This should be resolved with #1618. I'll test again once the new version for dev-tools is released.

g-awmalik:
Applied and verified.

Issue #1602, opened 2023-05-24
Ignore `/assets` folder in Cloud Build Triggers


NimJay:
### Background
* Each JSS has its own architectural diagram.
* [go/jss-diagrams](http://go/jss-diagrams) details the process for creating and maintaining these architectural diagrams.
* We would like to host the diagram source files (PNG, SVG, and even JSON) inside the Blueprint GitHub repo of the JSS.
* I am proposing that we use a root-level `/assets/` folder for these diagrams, inside each JSS's GitHub repo.
* The `/assets/` can also be used for any other images that are are used for repo documentation, etc.

### Problem
* Currently when we update the `/assets/` folder, the Terraform integration test for Cloud Build Trigger get invoked.
* These integration tests can take about 1 hour to complete.

### Solution
* The integration tests' Cloud Build Triggers should ignore changes to the `/assets/` folder.

### Additional info

* See this [pull-request for JSS 3.1](https://github.com/GoogleCloudPlatform/terraform-ml-image-annotation-gcf/pull/8).
* I have suggested updates to the [HSA/JSS TF Authoring guide](http://go/jss-tf-authoring).

NimJay:
This has been implemented via https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1609.

PR #1609, opened 2023-05-24
feat: Ignore assets/ in int_triggers

NimJay:
This implements #1602.


apeabody:
Would we ever want to run the INT for just images (`png`, `svg`, etc), or would it make sense to exclude those instead/as well?  If I recall, `json` is used under some /test/integration, and likely other places, so we wouldn't want to globally exclude it.

@bharathkkb @g-awmalik WDYT?

apeabody (changes requested):
Thanks @NimJay!

apeabody:
> Would we ever want to run the INT for just images (`png`, `svg`, etc), or would it make sense to exclude those instead/as well? If I recall, `json` is used under some /test/integration, and likely other places, so we wouldn't want to globally exclude it.
> 
> @bharathkkb @g-awmalik WDYT?

Per discussion let's leave as folder based for now, but add infra/assets/**

apeabody:
Hi @NimJay - Is this PR waiting on anything?  It's still marked as draft.

NimJay:
Hi @apeabody,
Thank you for the quick review and context (into your discussion with @bharathkkb)!

You asked:
>  Is this PR waiting on anything?

I was wondering if I need approval from either @bharathkkb or @g-awmalik
because of the [Google Doc comment thread/discussion here](https://docs.google.com/document/d/1EQjWLziWLGIMuQ1DtU3wPxhpGkrl_WVTdArOCJHr9ic/edit).
If you don't think so, feel free to merge. All looks good to me. :)

bharathkkb (approved):

apeabody (approved):

apeabody:
applied

PR #1616, opened 2023-05-25
chore: fix lint.yaml.tftpl

apeabody:
correct bug introduced when we switched from hard coded Makefile to find


bharathkkb (approved):

apeabody:
applied

PR #1618, opened 2023-05-27
feat: updates metadata validation to walk a path

g-awmalik:
Previously the validation logic would specifically look for metadata files in the current (or provided) path and in the modules/ dir underneath it. Now it walks the path provided regardless of the folder structure and finds/validates metadata files everywhere.

### DIFF
--- a/cli/bpmetadata/validate.go
+++ b/cli/bpmetadata/validate.go
@@ -4,7 +4,6 @@ import (
 	_ "embed"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"path"
 
 	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/cli/util"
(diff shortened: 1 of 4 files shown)


cli/util/file_test.go
```diff
@@ -53,3 +54,50 @@ func TestTFDirectories(t *testing.T) {
...
+		wantErr bool
+	}{
+		{
+			name:    "pattern for metadata files",
+			path:    "",
+			pattern: `.*/metadata(?:.display)?.yaml`,
+			want: []string{
+				"../testdata/bpmetadata/content/examples/acm/acm-terraform-blog-part1/terraform/metadata.yaml",
+				"../testdata/bpmetadata/content/examples/acm/metadata.display.yaml",
+				"../testdata/bpmetadata/content/examples/acm/metadata.yaml",
```
apeabody, line 73: Might want to add some test cases for "strange" situations:

"../testdata/bpmetadata/content/examples/acm/metadata.yaml.invalid",
"../testdata/bpmetadata/content/examples/acm/metadata.yaml/file",

cli/util/file_test.go
```diff
@@ -53,3 +54,50 @@ func TestTFDirectories(t *testing.T) {
...
+		name    string
+		path    string
+		pattern string
+		want    []string
+		wantErr bool
+	}{
+		{
+			name:    "pattern for metadata files",
+			path:    "",
+			pattern: `.*/metadata(?:.display)?.yaml`,
```
apeabody, line 69:
```suggestion
			pattern: `.*/metadata(?:.display)?.yaml$`,
```

cli/bpmetadata/validate.go
```diff
@@ -27,51 +26,18 @@ func validateMetadata(bpPath, wdPath string) error {
...
-	modulesPath := path.Join(bpPath, modulesPath)
-	_, err := os.Stat(modulesPath)
-	if err == nil {
-		subModuleDirs, err := util.WalkTerraformDirs(modulesPath)
-		if err != nil {
-			Log.Warn("unable to read the submodules i.e. modules/ folder", "err", err)
-		}
-
-		moduleDirs = append(moduleDirs, subModuleDirs...)
+	metadataFiles, err := util.FindFilesWithPattern(bpPath, `.*/metadata(?:.display)?.yaml`)
```
apeabody, line 29: Depends on the details, but do you need to escape the `/`?  `\/`

Probably want to anchor the end regardless:

```suggestion
	metadataFiles, err := util.FindFilesWithPattern(bpPath, `.*/metadata(?:.display)?.yaml$`)
```

apeabody (approved):
Hi @g-awmalik looks good - a few notes on the regex to avoid clearly edge cases.

cli/util/file_test.go
g-awmalik: Ack

cli/bpmetadata/validate.go
g-awmalik: Go doesn't requires you to escape if you provide the regex in back ticks. But anchoring it is valid. I'll do that.

cli/util/file_test.go
g-awmalik: Added.

PR #1628, opened 2023-05-31
feat: update to terraform-docs v0.16.0

apeabody:


bharathkkb:
Thanks @apeabody. I don't think we have automated tests for this. Were you able to validate in a repo?

apeabody:
> Thanks @apeabody. I don't think we have automated tests for this. Were you able to validate in a repo?

Yup - However there is a risk as this also switches to the new `inject` mode rather than using Perl.  So if there are any in-scope README.md files without `<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK --><!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->` the content will be added to the end of the file rather than the current no-op behavior.  However this might be desirable so we can identify/fix those README.md?

bharathkkb (approved):

PR #1655, opened 2023-06-07
chore: added owner in modules: dns, nat, router, log-export , vpn, vpc-sc

imrannayer:
Added owner in dns, nat, router, log-export , vpn, vpc-sc


g-awmalik (approved):

g-awmalik:
applied

PR #1657, opened 2023-06-08
chore: onboard secure notebook blueprint

gtsorbo:
legacy blueprint for AI notebook security is due for an update, and was never onboarded to CI properly


infra/terraform/test-org/org/locals.tf
```diff
@@ -663,5 +663,12 @@ locals {
       owners      = ["Saipriyavk", "ChrisBarefoot"]
       topics      = local.common_topics.net
     },
+    {
+      name        = "notebooks-blueprint-security"
```
g-awmalik, line 667: Any reason this is not prefixed with "terraform"?

gtsorbo: @g-awmalik This was an older blueprint and so it didn't follow standard naming - open to renaming the repo itself if that's possible

g-awmalik: Sorry for the delaying response. Let's let this one slide.

g-awmalik (approved):

g-awmalik (changes requested):
@gtsorbo - I tried applying but the name is too long. It's already 28 chars and with a prefix of "ci-" it becomes 31...should be less than 30. Can you rename this or provide an alternate `short_name` for this?

g-awmalik (approved):

g-awmalik:
Applied.

PR #1685, opened 2023-06-27
chore: enable scorecard

apeabody:


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@e38b1902ae4f44df626f11ba0734b14fb91f8f86 # v2.1.2
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          # (Optional) "write" PAT token. Uncomment the `repo_token` line below if:
+          # - you want to enable the Branch-Protection check on a *public* repository, or
```
bharathkkb, line 45: Does this apply to us since we have the `branch_protection_rule` trigger?

bharathkkb (approved):

.github/workflows/scorecard.yml
apeabody: Let me follow-up with your directly.

PR #1686, opened 2023-06-27
fix: escape period in regex

apeabody:


bharathkkb (approved):
Thanks @apeabody 

Could we also add a testcase for this like `git@githubAcom` doesn't match?

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/39838f58c18ae33c065243d028afd0d65c55bb34/cli/util/git_test.go#L41

apeabody:
> Thanks @apeabody
> 
> Could we also add a testcase for this like `git@githubAcom` doesn't match?
> 
> https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/39838f58c18ae33c065243d028afd0d65c55bb34/cli/util/git_test.go#L41

Sure - ironically just means the value is unchanged.

PR #1705, opened 2023-07-10
feat: adds processing logic for bpmetadata to use protobuf

g-awmalik:
In order to make metadata schema definition language independent, bpmetadata now processes metadata content using protobuf.

Known issues:
- BlueprintVariable.DefaultValue is changed from interface{} to string. This can cause error messages while running bpmetadata on existing blueprints. Solution: run bpmetadata with -f to force creation of proto compliant metadata.
- Go docs will most likely have superfluous types and methods not directly related to metadata. Solution: use protos as the source of truth for metadata definition and specs.


cli/Makefile
```diff
@@ -10,21 +10,33 @@ NAME=cft
...
+.PHONY: protoc-gen
+protoc-gen:
+	protoc -I=${SRC_PROTO_DIR} --go_opt=paths=source_relative --go_out=${PROTOC_DIR} ${SRC_PROTO_DIR}/*.proto
+	protoc-go-inject-tag -input="${PROTOC_DIR}/*.pb.go"
+
+.PHONY: protoc-deps
+protoc-deps:
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.2
+	go install github.com/favadi/protoc-go-inject-tag@v1.4.0
```
bharathkkb, line 39: nit: It would be better to just run protoc within a docker container. Maybe we can add it to our tools image?

cli/bpmetadata/schema/generate.go
```diff
@@ -41,6 +41,16 @@ func GenerateSchema() ([]byte, error) {
 	r := &jsonschema.Reflector{}
 	s := r.Reflect(&bpmetadata.BlueprintMetadata{})
 	s.Version = "http://json-schema.org/draft-07/schema#"
+
+	// defaultValue was defined as interface{} and has changed to
```
bharathkkb, line 45: Couldn't we keep it as Any?

https://protobuf.dev/programming-guides/proto3/#any

cli/bpmetadata/int-test/goldens/golden-metadata.yaml
```diff
@@ -29,133 +29,94 @@ spec:
...
...h to live and/or archived objects. Supported values include: \"LIVE\", \"ARCHIVED\", \"ANY\".\n    # - matches_storage_class - (Optional) Comma delimited string for storage class of objects to satisfy this condition. Supported values include: MULTI_REGIONAL, REGIONAL, NEARLINE, COLDLINE, STANDARD, DURABLE_REDUCED_AVAILABILITY.\n    # - num_newer_versions - (Optional) Relevant only for versioned objects. The number of newer versions of an object to satisfy this condition.\n    # - custom_time_before - (Optional) A date in the RFC 3339 format YYYY-MM-DD. This condition is satisfied when the customTime metadata for the object is set to an earlier date than the date used in this lifecycle condition.\n    # - days_since_custom_time - (Optional) The number of days from the Custom-Time metadata attribute after which this condition becomes true.\n    # - days_since_noncurrent_time - (Optional) Relevant only for versioned objects. Number of days elapsed since the noncurrent timestamp of an object.\n    # - noncurrent_time_before - (Optional) Relevant only for versioned objects. The date in RFC 3339 (e.g. 2017-06-13) when the object became nonconcurrent.\n    condition = map(string)\n  })))"
```
bharathkkb, line 47: This seems less readable than before. Is this due to pb -> json -> yaml?

cli/bpmetadata/tfconfig_test.go
```diff
@@ -68,22 +81,46 @@ func TestTFInterfaces(t *testing.T) {
...
-				Description:  tt.wantDescription,
-				DefaultValue: tt.wantDefault,
-				Required:     tt.wantRequired,
-				VarType:      tt.wantVarType,
-			})
+			i := slices.IndexFunc(got.Variables, func(v *BlueprintVariable) bool { return v.Name == tt.varName })
+			if got.Variables[i].Name != tt.varName {
+				t.Errorf("getBlueprintInterfaces() - Variable.Name = %v, want %v", got.Variables[i].Name, tt.varName)
+				return
+			}
```
bharathkkb, line 88: Couldn't we just assert.Equal() each of these assertions?

bharathkkb (commented):
Thanks @g-awmalik! Are there any benefits to checking in the generated files? I was thinking we could generate these at build time and add to gitignore?

cli/bpmetadata/schema/generate.go
g-awmalik: I don't think we can because marshaling/unmarshaling to/from Any can only be done with/to proto messages and not primitive types AFAIK. Here we have the default value being returned as an interface{} and we need to convert it to some valid proto type and string seems to be the most appropriate type.

Open to other suggestions if you have any.

cli/bpmetadata/int-test/goldens/golden-metadata.yaml
g-awmalik: Yeah the json to yaml conversion messes it up.

cli/bpmetadata/tfconfig_test.go
g-awmalik: I like checking each attribute since it's easier to tell where exactly the failures occur.

cli/bpmetadata/schema/generate.go
bharathkkb: I did some digging and it looks like the value type might be a good choice. We can also convert from interface{} to value easily using https://pkg.go.dev/google.golang.org/protobuf@v1.31.0/types/known/structpb#NewValue. WDYT?

cli/bpmetadata/int-test/goldens/golden-metadata.yaml
bharathkkb: Resolving since we chatted offline

cli/bpmetadata/tfconfig_test.go
bharathkkb: Resolving since we chatted offline

cli/Makefile
g-awmalik: I'm gonna do this in a follow-up PR

cli/bpmetadata/schema/generate.go
g-awmalik: Thanks a bunch. That worked beautifully.

bharathkkb (approved):
Can we also remove the generated files?

g-awmalik:
> Can we also remove the generated files?

We can but i'll do that after this goes through because the CFT CLI Tests check will fail since it executes `go test ./...`. We need to change that to `make go_test`. I do a separate PR for that while this is being reviewed by consumer teams.

cli/Makefile
g-awmalik: Done with #1713

PR #1712, opened 2023-07-14
feat: update dev-tools 1.14 to ruby v3

apeabody:


bharathkkb (approved):
Thanks @apeabody! Any breaking changes we should be aware of with 3.0?

apeabody:
> Thanks @apeabody! Any breaking changes we should be aware of with 3.0?

I believe Ruby is primarily used for legacy Kitchen tests.  We already [upgraded](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/commit/af3fb816d423e0cf420afa6a9de17fe09afadb12) to kitchen-terraform v6 in dev-tools 1.12 which [added](https://github.com/newcontext-oss/kitchen-terraform/blob/main/CHANGELOG.md#600---2021-07-07) support for Ruby v3.

That said, this bumps the dev-tools minor release, so any impact would be limited to repos pinned to only the major version, which could always be switch to tracking the minor if needed.  Long term this is to unblock updating the Alpine release, but I left that for a future minor version to limit the changes at one time.

PR #1715, opened 2023-07-18
feat: adds protoc to devtools image

g-awmalik:


infra/build/developer-tools/build/install_protoc.sh
```diff
@@ -0,0 +1,33 @@
+#! /bin/bash
+# Copyright 2021 Google LLC
```
apeabody, line 2:
```suggestion
# Copyright 2023 Google LLC
```

infra/build/developer-tools/build/install_protoc.sh
```diff
@@ -0,0 +1,35 @@
...
+PROTOC_GEN_GO_INJECT_TAG=$4
+
+cd /build
+
+curl -LO "https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-linux-x86_64.zip"
+unzip "protoc-${PROTOC_VERSION}-linux-x86_64.zip" -d $HOME/.local
+rm "protoc-${PROTOC_VERSION}-linux-x86_64.zip"
+chmod 755 $HOME/.local/bin/protoc
+cp $HOME/.local/bin/protoc /usr/local/bin/
+cp -R $HOME/.local/include/* /usr/local/include/
```
bharathkkb, line 31: nit: Can we remove `$HOME/.local/` since everything is copied?

bharathkkb (approved):

infra/build/developer-tools/build/install_protoc.sh
g-awmalik: I think it's better to keep it since there may be other things under `/build/bin` and/or `/build/include` from other installations.

PR #1725, opened 2023-07-24
fix: Update task_helper_functions.sh to omit pb and index files

ivanmkc:
See https://github.com/HSA-Integration/Annotate-images-with-Vertex-AI/pull/92, which uses pb and index files.


bharathkkb (commented):
To skip everything, we should also change in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/1b65eb48b963317233b0c78634463a4700d5af29/infra/build/developer-tools/build/scripts/task_helper_functions.sh#L247

https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/1b65eb48b963317233b0c78634463a4700d5af29/infra/build/developer-tools/build/scripts/task_helper_functions.sh#L68

bharathkkb (approved):

Issue #1726, opened 2023-07-25
AssertHTTP: WithRetry methods fail tests on first failed attempt


grayside:
In #1707, I wrote a new set of HTTP assert helpers to facilitate testing web services as part of blueprint testing.

After some additional experimentation on the naive approach to the WithRetry methods, I've realized that even though we retry, the first failed HTTP request will fail the tests, because we call `t.Error` inside the retry loop.

Instead, we need to catch the last failure/error of the retry loop and use that to call `t.Error`.

grayside:
I am working on an implementation where the last error returned from Poll is used as the error on which to assert in the Assert methods.

As part of this, I am also fixing:
* Use of `utils.GetLoggerFromT`
* Improved return from httpResponse on empty conditions
* Expanded test coverage (catching some additional edge cases)
* Configurable retries, because the current implementation would not have helped with what I needed to do on Monday.

At this point, I would have considered dropping the `AssertXWithRetry` methods, but since they are already there, I've kept them in this PR.

PR #1729, opened 2023-07-26
feat: add utils.PollE

grayside:
This is part one of fixing #1726, splitting `utils.Poll` into two functions, one that gives the caller insights to decide how to handle failure, and the other preserving existing behavior.

Additional, this PR adds initial test coverage to Polling behavior. If the iteration count were not loop-scoped, I might have made this `PollE(...) (int, error)` and returned the iteration count as a first value.


grayside:
Fixed now, ready for review.

infra/blueprint-test/pkg/utils/poll.go
```diff
@@ -45,10 +55,12 @@ func Poll(t testing.TB, condition func() (bool, error), numRetries int, interval
 	}
 
 	if err != nil {
-		t.Fatalf("failed to pull provided condition after %d retries, last error: %v", numRetries, err)
+		return fmt.Errorf("failed to pull provided condition after %d retries, last error: %v", numRetries, err)
```
bharathkkb, line 58: nit: would custom errors be beneficial for users distinguish between errors types?

bharathkkb (approved):
Thanks for adding tests too! Looks like golint is complaining about something but LGTM otherwise

infra/blueprint-test/pkg/utils/poll.go
grayside: Added.

PR #1761, opened 2023-08-11
chore: update golangci-lint

apeabody:


g-awmalik (approved):
Is not being handle renovate?

apeabody:
> Is not being handle renovate?

Not currently, we would need to see if we can design a regex manager, easy if we breakout the version similar to dev-tools.

PR #1768, opened 2023-08-15
feat: add disk image reference in machine type property extension

rzy-goog:
Create a new variable under GCEMachineTypeExtension to reference the disk image property, so that we can gather data on the image, like architecture. 

Will help for new ARM support for marketplace products.


g-awmalik (approved):
LGTM. There's a change we need to make on the g3 side before we can merge this. I'll cc you on the CL.

PR #1773, opened 2023-08-16
fix: exclude parquet files from checks

nevzheng:
I am drafting a PR to update terraform samples in https://github.com/terraform-google-modules/terraform-docs-samples/pull/474

The lint stage of the CI is failing as we're attempting to check that parquet files end in new lines. You can't append new lines to parquet files wihtout invalidating them.

So we should not check that parquet files end in newlines.

Related PR: (https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1725)


apeabody (changes requested):
Thanks for the contribution @nevzheng!

To be safe given all the formatting, we should also bump the dev-tools patch version (e.g. `1.14.4`https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/Makefile#L56

infra/build/developer-tools/build/scripts/task_helper_functions.sh
```diff
@@ -364,30 +365,30 @@ function check_tflint() {
...
-      # load default ruleset
-      tflintCfg="/root/tflint/.tflint.example.hcl"
-      # if module, load tighter ruleset
+      echo "Skipping ${path}"
+      continue
+    fi
+    # load default ruleset
+    tflintCfg="/root/tflint/.tflint.example.hcl"
+    # if module, load tighter ruleset
       if [[ $path == "." || $path == "./modules"* ]];then
```
apeabody: I think this might need to move up a level?

nevzheng:
moving to attempt 2 github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1776

PR #1789, opened 2023-08-24
Revert "chore: update update-tooling.yml"

bharathkkb:
Reverts GoogleCloudPlatform/cloud-foundation-toolkit#1757 since usage is only for developer tools/test


apeabody:
@bharathkkb - I'll open a quick follow-up as I still want to bump the GK verison.

apeabody (approved):

PR #1792, opened 2023-08-24
chore: reduce number of dev-tools layers

apeabody:
* Reduce number of image layers for automated build reliability
* Consolidate install of `apk` and `gcloud` module components for speed and efficiency
* Use distro's python modules rather than pip
* Move `cookiecutter` to just krm
* Increase Ruby module download retry for build reliability


infra/build/developer-tools-krm/Dockerfile
```diff
@@ -64,3 +64,7 @@ RUN rm -rf /build
 # Add dind helper for prow
 ADD ./build/scripts/runner.sh /usr/local/bin/
 ADD ./build/scripts/prow_entrypoint.sh /usr/local/bin/
+
+# cookiecutter is used for terraform-google-module-template tests
+# https://github.com/terraform-google-modules/terraform-google-module-template
+RUN pip install cookiecutter
```
bharathkkb, line 70: Do we need this since this is only used in KRM tests?

bharathkkb (approved):
Thanks for fixing this @apeabody!

infra/build/developer-tools-krm/Dockerfile
apeabody: For now I moved to just the krm image as that image is used for github.com/terraform-google-modules/terraform-google-module-template CI

A future option would be to do something specific in the repo so this could be removed entirely.

Issue #1732, opened 2023-07-27
Dev-tools minor update with: Ruby 3.2 (including bundler), Kitchen-terraform 7, and Alpine 3.18


apeabody:

apeabody:
Let's first complete https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1633

Issue #1547, opened 2023-05-12
Update dev-tools Alpine


apeabody:
We also need to update `dev-tools` to Ruby 3, https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1553 should address most Ruby dependencies.

PR #1801, opened 2023-09-01
chore: update dev-tools v1.15 on Alpine 3.18

apeabody:
This is the update we've been preparing for!
* Update to Alpine 3.18
* Use Alpine's Go (still 1.20) and Ruby
* Ruby 3.3
* kitchen-terraform 7.0

Fix: #1732
Fix: #1547


bharathkkb (approved):
Thanks @apeabody!

Issue #1738, opened 2023-07-31
Remove autogen files produced by protoc


g-awmalik:
Currently, 3 files are generated by the protoc tool which shouldn't ideally be checked in.

Limitation: the runner used to perform unit tests for macOS doesn't support docker which limits our ability to generate these files as part of the CI process.

PR #1815, opened 2023-09-06
chore: use cli full make test suite

apeabody:
Use the full cli's `make test` rather than just `go test`.  Unfortunately the Mac runner doesn't include `docker`.


apeabody:
@g-awmalik If we need to test on Mac, we can still include a separate `go test`, however I think the value is somewhat limited.  Dropping completely would also unblock https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/1738

g-awmalik (approved):

PR #1819, opened 2023-09-07
chore: dev-tools 1.16 update kustomize, terragrunt, tflint, remove terraform-validator

apeabody:


g-awmalik:
Please hold off on release this. I have a change in the CLI that I'd like to push in here. Thanks.

infra/build/developer-tools/Dockerfile
```diff
@@ -89,7 +88,6 @@ RUN /build/install_shellcheck.sh && \
     /build/install_hadolint.sh && \
     /build/install_addlicense.sh && \
     /build/install_gsuite_terraform_provider.sh ${GSUITE_PROVIDER_VERSION} && \
-    /build/install_terraform_validator.sh ${TERRAFORM_VALIDATOR_VERSION} && \
```
g-awmalik, line 92: Any reason why we're removing this? I think example foundation still uses it.

apeabody: Technically it is archived: https://github.com/GoogleCloudPlatform/terraform-validator

It looks like `terraform-example-foundation` has moved to `gcloud beta terraform vet` https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/build/tf-wrapper.sh#L155 and repos using the wrapper script will now use `gcloud beta terraform vet` as well.

If there is another repo which uses `terraform-validator` directly, we can always keep it on `1.15` till it's updated directly.

g-awmalik (approved):

PR #1821, opened 2023-09-08
fix: adds blueprint metadata path to errors

g-awmalik:
Couple of fixes:
- Better error messaging with path for the module
- Marshaling of BlueprintMetadata directly from the type instead of converting it to JSON and then YAML. The library used adds a indent for list types but I think that's OK given the formatting is much better now.


apeabody (approved):
LGTM - I actually prefer the list indent.

PR #1878, opened 2023-10-12
feat: adding support for bq

bradmiro:
Adding support for `bq` command line tool.


infra/blueprint-test/pkg/bq/bq.go
```diff
@@ -0,0 +1,165 @@
...
+		return fmt.Sprintf("%+v", msg)
+	}
+	if len(msgAndArgs) > 1 {
+		return fmt.Sprintf(msgAndArgs[0].(string), msgAndArgs[1:]...)
+	}
+	return ""
+}
+
+// ActivateCredsAndEnvVars activates credentials and exports auth related envvars.
+func ActivateCredsAndEnvVars(t testing.TB, creds string) {
```
bharathkkb, line 152: This does not seem BQ specific and might be a better fit in utils?

infra/blueprint-test/pkg/bq/bq.go
```diff
@@ -0,0 +1,165 @@
...
+//
+// Runf(t, "ls --datasets --project_id=%s", "projectId")
+//
+// It fails the test if there are any errors executing the bq command or parsing the output value.
+func Runf(t testing.TB, cmd string, args ...interface{}) gjson.Result {
+	return Run(t, stringFromTextAndArgs(append([]interface{}{cmd}, args...)...))
+}
+
+// stringFromTextAndArgs convert msg and args to formatted text
+func stringFromTextAndArgs(msgAndArgs ...interface{}) string {
```
bharathkkb, line 134: Since we are starting to use this for both gcloud and bq, I think centralizing this in utils might be good.

bharathkkb (commented):
Thanks for the PR @bradmiro

infra/blueprint-test/pkg/bq/bq.go
bradmiro: I copied this directly from `gcloud.go` since auth is the same - I can move it from here (and there) to utils if that makes sense. Or I can leave it as-is in `gcloud.go`

infra/blueprint-test/pkg/bq/bq.go
bradmiro: I can move this out of both files but that might result in a breaking change if the function in `gcloud.go` was referenced in any downstream projects - wdyt?

bharathkkb: Since it was unexported from gcloud package it is okay to move

bradmiro: Works for me.

bradmiro: Done!

bharathkkb (approved):

PR #1955, opened 2023-11-08
fix: updates TF versions for dev & test org configs

g-awmalik:


apeabody (approved):
The tests don't really exercise these, but it should be fine if they are mostly wrapping modules.

PR #2000, opened 2023-12-04
fix: add bq init step to create local config file

bradmiro:
If the local environment does not have a .bigqueryrc file, using the tool will faill as bq initializing itself does not output valid json data.

### DIFF
--- a/infra/blueprint-test/pkg/bq/bq.go
+++ b/infra/blueprint-test/pkg/bq/bq.go
@@ -18,6 +18,7 @@
 package bq
 
 import (
+	"os"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/infra/blueprint-test/pkg/utils"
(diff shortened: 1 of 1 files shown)


infra/blueprint-test/pkg/bq/bq.go
```diff
@@ -91,6 +110,11 @@ func RunCmdE(t testing.TB, cmd string, opts ...cmdOption) (string, error) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	err = initBq()
```
bharathkkb, line 114: nit: Could we pass in t to initBq if we are planning to call t.Fatal(err) anyway?

infra/blueprint-test/pkg/bq/bq.go
```diff
@@ -76,6 +77,24 @@ func newCmdConfig(opts ...cmdOption) (*CmdCfg, error) {
...
 
+// initBq checks for a local .bigqueryrc file and creates an empty one if not to avoid forced bigquery initialization, which doesn't output valid json.
+func initBq() (error) {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+	fileName := homeDir + "/.bigqueryrc"
+	 _ , err = os.Stat(fileName)
+	if os.IsNotExist(err) {
```
bharathkkb, line 88: nit: I think we could keep this more left aligned like 
```go
if !os.IsNotExist(err) { return nil }
file, err ...
```

bharathkkb (approved):
Thanks for the PR @bradmiro, a few nits

infra/blueprint-test/pkg/bq/bq.go
bradmiro: Ack

infra/blueprint-test/pkg/bq/bq.go
bradmiro: Since `initBq` doesn't currently call `t.Fatalf` inside of it (instead returning an error value), what would passing t accomplish?

bharathkkb: I was suggesting `initBq` could call t.Fatalf by passing `t` if we are going to call t.Fatal(err) at the call site anyway.

bradmiro: Ack, fixed.

bharathkkb (approved):

PR #2048, opened 2023-12-18
fix: omit null defaultValue in bpmetadata

rzy-goog:
Related to b/316180349

Currently, cft blueprint metadata -d appears to set the defaultValue to null if there's no default set in the Terraform file.

This breaks current terraform metadata parsing in g3

This change will omit defaultvalue from being added when null.


cli/bpmetadata/tfconfig.go
```diff
@@ -231,12 +231,12 @@ func getBlueprintVariable(modVar *tfconfig.Variable) *BlueprintVariable {
...
 	}
-
-	vl, err := structpb.NewValue(modVar.Default)
-	if err == nil {
-		v.DefaultValue = vl
+	if modVar.Default != nil {
+		vl, err := structpb.NewValue(modVar.Default)
+		if err == nil {
+			v.DefaultValue = vl
+		}
```
g-awmalik, line 238: nit: can we make this a bit more left aligned?

```suggestion
	if modVar.Default == nil {
	  return v	  
	}
        
        vl, err := structpb.NewValue(modVar.Default)
        if err == nil {
	  v.DefaultValue = vl
        }
        
        return v
```

g-awmalik (approved):
LGTM except for a minor nit.

g-awmalik:
Also, please update the version with a patch: https://github.com/rzy-goog/cloud-foundation-toolkit/blob/master/cli/Makefile#L4

g-awmalik (changes requested):
Please update the version with a patch: https://github.com/rzy-goog/cloud-foundation-toolkit/blob/master/cli/Makefile#L4

g-awmalik (approved):

PR #2056, opened 2023-12-20
feat: module swapper hcl source targeting

bharathkkb:
This upgrades module swapper from simple string matching for source swaps to a better approach which parses HCL. This approach makes it easy to restore sources without restore markers and was used  to create the automated PRs. Hoping in a future iteration we can expose this as a lib for usage in tft to automatically swap and un swap when running tests.

Versions still relies on string matching and restore markers since hclwrite does not seem to have a good way to insert attribs at a specific position. Also some minor enhancements to support ssh remotes.


infra/module-swapper/cmd/swap.go
```diff
@@ -54,28 +66,30 @@ func getModuleNameRegistry(dir string) (string, string, error) {
...
-		return "", "", fmt.Errorf("Expected GitHub org/owner of form ModuleRegistry/ModuleRepo. Got: %s", namePrefix)
+	split := strings.Split(orgRepo, "/")
+	if len(split) != 2 {
+		return "", "", fmt.Errorf("expected GitHub remote of form https://github.com/ModuleRegistry/ModuleRepo, got: %s", remote)
 	}
-	moduleRegistry := namePrefix[:strings.LastIndex(namePrefix, "/")]
-	repoName := namePrefix[strings.LastIndex(namePrefix, "/")+1:]
+	org, repoName := split[0], split[1]
 
 	// module repos are prefixed with terraform-google-
```
apeabody, line 86: Do we want the option to use with non-confirming repo names?  https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/test-org/org/locals.tf#L120

infra/module-swapper/cmd/swap.go
```diff
@@ -54,28 +66,30 @@ func getModuleNameRegistry(dir string) (string, string, error) {
...
-	namePrefix := strings.ReplaceAll(remote, "https://github.com/", "")
-	if !strings.Contains(namePrefix, "/") {
-		return "", "", fmt.Errorf("Expected GitHub org/owner of form ModuleRegistry/ModuleRepo. Got: %s", namePrefix)
+	split := strings.Split(orgRepo, "/")
+	if len(split) != 2 {
+		return "", "", fmt.Errorf("expected GitHub remote of form https://github.com/ModuleRegistry/ModuleRepo, got: %s", remote)
 	}
-	moduleRegistry := namePrefix[:strings.LastIndex(namePrefix, "/")]
-	repoName := namePrefix[strings.LastIndex(namePrefix, "/")+1:]
+	org, repoName := split[0], split[1]
```
apeabody, line 84: Would it be valuable to include a check on valid `org` values?  e.g. `GoogleCloudPlatform` `terraform-google-modules`?

apeabody (approved):
Thanks @bharathkkb!

infra/module-swapper/cmd/swap.go
apeabody: Just saw these:
- https://github.com/terraform-google-modules/terraform-example-foundation/pull/1064
- https://github.com/terraform-google-modules/terraform-example-foundation/commit/bed56d961ba45dcf36a60d2ebc5e55fad5b3389e#commitcomment-135386931

apeabody: Sounds like at a minimum we might want to instead exit gracefully and log that the repo is not supported, etc.

bharathkkb: Yup! Fixed this

infra/module-swapper/cmd/swap.go
bharathkkb: Hmm I had added support for a `moduleRegistryPrefix` so users could explicitly specify a different org instead. Did we have a usecase were restricting is better?

apeabody: Ahh - I see your plan.  I was thinking it might be valuable to include those orgs as the "defaults" and permit overriding with the flag, but this direction is also fine.

PR #2067, opened 2023-12-27
chore(deps): bump test-org for TPG v5

apeabody:
tested: `terraform plan`


infra/terraform/test-org/org/versions.tf
```diff
@@ -18,23 +18,32 @@ terraform {
...
+      source  = "hashicorp/random"
+      version = ">= 2.3.1, < 4"
     }
     github = {
       source  = "integrations/github"
       version = "~> 5.0"
     }
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = ">= 2.13"
```
bharathkkb, line 46: Should we also consider constraining major version?

bharathkkb (approved):

infra/terraform/test-org/org/versions.tf
apeabody: Probably a good idea.  Done.

apeabody: no change to lock file

PR #2074, opened 2024-01-02
chore: correct stale pr exemptions

apeabody:
- `dependencies` is for prs, not issues
- We should also exempt pending releases


g-awmalik (approved):

apeabody:
applied

Issue #2060, opened 2023-12-26
Integrate module swapper with lint tests


bharathkkb:
It would be nice to automatically perform swaps as needed when doing lint tests

PR #2088, opened 2024-01-11
feat: add module swapper to lint script

bharathkkb:
fixes https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/2060

Also simplified the lint script since we no longer use the GH bot + markdown integration.


apeabody (approved):
Thanks @bharathkkb!

Any conflict with the lint workflow also running module-swapper?

bharathkkb:
@apeabody it should be a no-op if both are run

PR #2103, opened 2024-01-25
chore(TF): reduce number of centralized labels

apeabody:
- Reduce to a minimum number of mandatory centralized labels
- IMPORTANT: Rather than removing existing labels, they must be removed only from TF state


infra/terraform/test-org/github/labels.tf
```diff
@@ -26,96 +26,21 @@ locals {
       color : "d73a4a"
       description : "Something isn't working"
     },
-    {
-      name : "duplicate",
-      color : "cfd3d7"
-      description : "This issue or pull request already exists"
-    },
-    {
-      name : "good first issue",
```
bharathkkb, line 35: This might be good to keep to provide new contributors with easy tasks.

infra/terraform/test-org/github/labels.tf
```diff
@@ -26,96 +26,21 @@ locals {
...
-      color : "008672",
-      description : "Extra attention is needed"
-    },
-    {
-      name : "invalid",
-      color : "e4e669",
-      description : "Something doesn't seem right"
-    },
     {
       name : "question",
```
bharathkkb, line 30: I don't think this provides too much value

bharathkkb (approved):
Left a few comments but non blocking

infra/terraform/test-org/github/labels.tf
apeabody: SG - added back

infra/terraform/test-org/github/labels.tf
apeabody: SG - removed

apeabody:
applied - as the actual resources weren't removed they could always be re-added/imported if needed

PR #2147, opened 2024-02-13
chore: update golangci-lint renovate

apeabody:


g-awmalik:
This seems to be failing for config-validator

apeabody:
> This seems to be failing for config-validator

Yeah, I missed a comma.  But good to validate that the `renovate-config-validator` works! :)

g-awmalik (approved):

PR #2160, opened 2024-02-16
chore: add periodic support for docs-samples

apeabody:
- `_PERIODIC` env is to enable periodic specific behavior.  For example here: https://github.com/terraform-google-modules/terraform-docs-samples/blob/main/build/int.cloudbuild.yaml#L17
- `workflow_dispatch:` enables manual trigger of `lint`.  Useful to start testing a similar scheduled workflow.

Ref: https://github.com/terraform-google-modules/terraform-docs-samples/issues/520


bharathkkb (approved):

apeabody:
Hi @glasnt - Wanted to give you a heads up. This will expand the ability to run periodic INT tests to repos like `terraform-docs-samples`.  However the only "visible" change would be an issue opened with the periodic test status.

FYI: For `terraform-docs-samples` in particular I've added a `_PERIODIC = true` ENV so it's possible to later update the pruner to skip pruning on periodic tests so all of them get run.

glasnt:
> FYI: For `terraform-docs-samples` in particular I've added a `_PERIODIC = true` ENV so it's possible to later update the pruner to skip pruning on periodic tests so all of them get run.

Amazing! Thank you so much!

apeabody:
applied - I'll follow-up with the periodic reporter issue workflow after the trigger has a chance to run successfully a few times.

PR #2165, opened 2024-02-20
fix: adds deprecated tag to all version tags for dev-tools*

g-awmalik:
The vulnerability scanner expects just as many deprecated tags as the public image tags for it to skip scanning. Since we tag dev-tools with major, minor and patch, it requires three deprecated tags.


bharathkkb (approved):

bharathkkb:
For a future PR: I think this tagging section is big enough now that we might want to split this into a script for maintainability

PR #2251, opened 2024-04-04
fix: minor description and topic updates for page refresh

bharathkkb:


bharathkkb:
This has been applied

PR #2259, opened 2024-04-10
chore: bump go version

bharathkkb:


go.work
```diff
@@ -1,4 +1,4 @@
-go 1.21
+go 1.22
```
apeabody, line 1: The dev-tools image currently includes GO 1.21

bharathkkb: Ah thanks for catching! I guess we are using alpine go which still seems to be 1.21. This is not urgent and we can hold off-  I was just experimenting with something that needed 1.22 so I split this off into a PR.

apeabody: Yeah - The current Alpine v3.19 has GO v1.21.  Looks like Alpine v3.20 is targeted for ~June.

apeabody: Go 1.22 included in dev-tools 1.21

apeabody (approved):
LGTM

PR #2267, opened 2024-04-16
chore: update ci_teams to write for cloudbuild trigger

apeabody:
- Collaborator require `write`, `triage` is mapped to `read`: https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#add-a-repository-collaborator


infra/terraform/modules/repositories/main.tf
```diff
@@ -103,5 +103,5 @@ resource "github_team_repository" "ci_teams" {
   }
   repository = each.value.repo
   team_id    = each.value.team
-  permission = "triage"
+  permission = "write"
```
bharathkkb, line 106: Although they have write IIUC they won't be able to push any changes directly due to branch protection?

bharathkkb (approved):

infra/terraform/modules/repositories/main.tf
apeabody: Correct - CODEOWNERS for the main branch, which `write` can't override.

g-awmalik:
Applied

PR #2277, opened 2024-04-19
chore: add release automation for tflint plugin

bharathkkb:
This adds release please configs for releasing plugin and a workflow that is triggered on release to build and upload artifacts.  

Goreleaser is used to build and package the artifacts in the format expected by tflint https://github.com/terraform-linters/tflint/blob/master/docs/developer-guide/plugins.md#4-creating-a-github-release. We don't use goreleaser to publish them as mono support is a pro (paid) feature https://goreleaser.com/customization/monorepo. Instead I used the gh cli to upload artifacts to the release.

### DIFF
--- a/.github/workflows/release-tflint-plugin.yml
+++ b/.github/workflows/release-tflint-plugin.yml
@@ -1,0 +1,39 @@
+name: release-tflint
+
+on:
+  push:
+    tags:
+    - tflint-ruleset-blueprint/v*.*.*
+
+permissions:
+  contents: write
+  id-token: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: 'tflint-ruleset-blueprint'
+    steps:
+    - name: Checkout
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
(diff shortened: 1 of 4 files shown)


.github/workflows/release-tflint-plugin.yml
```diff
@@ -0,0 +1,39 @@
...
+    - run: echo "${{env.GORELEASER_CURRENT_TAG}}"
+    - name: Run GoReleaser
+      uses: goreleaser/goreleaser-action@v5
+      with:
+        version: latest
+        args: release --clean --skip=validate,publish
+        workdir: tflint-ruleset-blueprint
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    - name: upload
```
apeabody, line 35: @bharathkkb Could we include a "test" release-tflint-plugin workflow which does everything other than the final upload step? That way we can at least confirm through the `Run GoReleaser` process works for Pull Requests (such as this one).

apeabody (approved):

.github/workflows/release-tflint-plugin.yml
bharathkkb: Added a goreleaser test workflow

PR #2329, opened 2024-05-07
chore(CI): cache lint tf provider cache

apeabody:
PoC for https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/2327


.github/workflows/lint.yaml
```diff
@@ -32,6 +32,14 @@ jobs:
     runs-on: 'ubuntu-latest'
     steps:
       - uses: 'actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b' # v4.1.4
+      - name: Cache lint-infra-terraform
+        id: cache-lint-infra-terraform
+        uses: actions/cache@v4
```
bharathkkb, line 37: Should we pin like other actions?

.github/workflows/lint.yaml
```diff
@@ -32,6 +32,14 @@ jobs:
...
+        with:
+          path: |
+            ${{ github.workspace }}/test/integration/tmp/.terraform
+          key: cache-lint-infra-terraform-${{ github.run_id }}
```
bharathkkb, line 41: Wondering if using a lockfile hash would allow for potentially wider matching across runs (although IIUC with restore-key we will fall back anyway to a cache-lint-infra-terraform* cache )

https://github.com/actions/cache/blob/main/caching-strategies.md#updating-cache-for-any-change-in-the-dependencies

bharathkkb (approved):
Thanks Andrew! This should be a nice boost

bharathkkb:
Also wondering could we use this for caching our docker image layers?

apeabody:
> Also wondering could we use this for caching our docker image layers?

I was also testing that, but the save/restore from cache was actually slower.  That said more research is warranted.

.github/workflows/lint.yaml
apeabody: Yes! :)

.github/workflows/lint.yaml
apeabody: Yeah, I was hoping to further refine this technique before wide spread deployment (https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/2327).  As you guessed this is because it's not possible to update an existing cache, so currently it always creates a "new" cache during clean-up to capture any potential changes.  I need to poke around in `.terraform` to see if there is a suitable file, etc.

PR #2364, opened 2024-05-17
fix(tft): remove default retry of inconsistent provider

apeabody:


infra/blueprint-test/pkg/tft/terraform.go
```diff
@@ -309,6 +309,9 @@ func (b *TFBlueprintTest) GetTFOptions() *terraform.Options {
 		RetryableTerraformErrors: b.retryableTerraformErrors,
 		NoColor:                  true,
 	})
+	// Remove Default "rovider produced inconsistent result after apply" from RetryableTerraformErrors
+	// Retrying these errors generally results in confusing already exist errors
+	delete(newOptions.RetryableTerraformErrors, ".*Provider produced inconsistent result after apply.*")
```
bharathkkb, line 314: Couldn't we remove these at the call sites?

apeabody: This removal is for one that is pulled in with a larger list from the ealier `terraform.WithDefaultRetryableErrors` call: https://github.com/gruntwork-io/terratest/blob/master/modules/terraform/options.go#L14

bharathkkb: Got it thank you!

bharathkkb (approved):

apeabody:
Going to leave this out for now.  If we decide to add we should make if configurable.

PR #2367, opened 2024-05-20
chore: enable parallel lint on terraform-docs-samples

apeabody:


apeabody:
Marking as draft pending @glasnt feedback

infra/terraform/test-org/org/locals.tf
```diff
@@ -543,6 +543,10 @@ locals {
       org             = "terraform-google-modules"
       module          = false
       enable_periodic = true
+      lint_env = {
+        "ENABLE_PARALLEL"     = "1",
```
glasnt, line 547: LGTM 👍

apeabody: Turns out terraform-docs-samples uses a forked config, so we'll use https://github.com/terraform-google-modules/terraform-docs-samples/pull/672

PR #2369, opened 2024-05-21
chore(CI): add tfenv lint-terraform-min-required

apeabody:


bharathkkb:
Thanks for exploring this @apeabody! I am wondering if it is simpler to start with a centralized min version enforcement at a specific version and then allow particular folks to override if needed from [onboarding config](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/test-org/org/locals.tf). 

We do lose the ability to automatically adjust based on config in repo but having that centralized control would be nice so updating min version has to go through this config.

apeabody:
> Thanks for exploring this @apeabody! I am wondering if it is simpler to start with a centralized min version enforcement at a specific version and then allow particular folks to override if needed from [onboarding config](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/test-org/org/locals.tf).
> 
> We do lose the ability to automatically adjust based on config in repo but having that centralized control would be nice so updating min version has to go through this config.

Sure, let's discuss further offline as I suspect for this alternative implementation we'd likely need to also/first bump the minimum to a common TF v1.3 (many are still v0.13) across the repos.  We might be able to use renovate and/or tflint to help with that aspect.

PR #2440, opened 2024-07-01
feat: Add blueprint connections and related metadata

bharathkkb:


cli/bpmetadata/proto/bpmetadata.proto
```diff
@@ -453,6 +468,43 @@ message BlueprintVariable {
   string var_type = 3; // @gotags: json:"varType,omitempty" yaml:"varType,omitempty"
   google.protobuf.Value default_value = 4; // @gotags: json:"defaultValue,omitempty" yaml:"defaultValue,omitempty"
   bool required = 5; // @gotags: json:"required,omitempty" yaml:"required,omitempty"
+  // Incoming connections to this variable.
```
g-awmalik, line 471: Can we define "connections" a bit more explicitly so it's easier for a someone without ADC context to understand?

cli/bpmetadata/proto/bpmetadata_ui.proto
```diff
@@ -126,6 +127,22 @@ message DisplayVariable {
   // Optional.
   // Gen: manually-authored
   string boolean_group = 20; // @gotags: json:"booleanGroup,omitempty" yaml:"booleanGroup,omitempty"
+
+  // Alternate default value.
```
g-awmalik, line 131: What is this used for?

g-awmalik (commented):
LGTM for the most part but a few concepts require a bit more description.

g-awmalik:
@njlisi - please review.

cli/bpmetadata/proto/bpmetadata.proto
bharathkkb: good point, added some more details

cli/bpmetadata/proto/bpmetadata_ui.proto
bharathkkb: Added a line with expected usage.

cli/bpmetadata/proto/bpmetadata_ui.proto
```diff
@@ -126,6 +127,23 @@ message DisplayVariable {
...
+  // This allows authors to define an alternative value for pre identified usecases such as security.
```
njlisi, line 132: Could you add some more context here? "AlternateDefault" name gives an impression that this field will provide a different value in place of the _default_value_ field in _BlueprintVariable_(bpmetadata.proto), is that correct? Is it used for providing supporting information for the default_value?

njlisi (commented):
LGTM

cli/bpmetadata/proto/bpmetadata_ui.proto
```diff
@@ -126,6 +127,23 @@ message DisplayVariable {
...
+  // Gen: manually-authored.
+  message AlternateDefault {
+    // Type of the alternate default.
+    enum AlternateType {
```
q2w, line 136: Going forward if a new alternate type is needed, will we update this enum? Let's say if we want a default for ADC use case, would we created new alternate type?

q2w (approved):

cli/bpmetadata/proto/bpmetadata_ui.proto
bharathkkb: Added some more info

> "AlternateDefault" name gives an impression that this field will provide a different value in place of the default_value field in BlueprintVariable(bpmetadata.proto), is that correct?

Yup! Its intended to be a hand authored alternate to the one in BlueprintVariable.

cli/bpmetadata/proto/bpmetadata_ui.proto
bharathkkb: Yes and I am thinking ADC can also able to leverage or surface some these existing alternates

g-awmalik (approved):

PR #2445, opened 2024-07-09
fix: Disable JSS 15.1 periodic tests

NimJay:
* See Google-internal bug: [b/351840017](http://b/351840017)
* We would like to stop all deployments of JSS 15.1.
* JSS 15.1 = Deploy a Java application with Compute Engine
* CC: @donmccasland, @duncantech


g-awmalik (approved):

g-awmalik:
Applied

PR #2496, opened 2024-08-07
feat: add provider version to metadata generation

qz267:
Feature request from: b/356844884

Add [ProviderVersion](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/97922719e7e251f611535079e1969fbb4af6bf9b/cli/bpmetadata/proto/bpmetadata.proto#L202) to metadata auto generation. Changes including:
-   Added `parseBlueprintProviderVersions` function to `tfconfig.go`
-   Added `TestTFProviderVersion` to `tfconfig_test.go`
-   Wired up `parseBlueprintProviderVersions` to `getBlueprintRequirements` function, which was called in the metadata generation process.
-   Update [golden-metadata.yaml](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/97922719e7e251f611535079e1969fbb4af6bf9b/cli/bpmetadata/int-test/goldens/golden-metadata.yaml#L240-L251) `requirements` block, added provider versions block.

Checks:
-   `make go_test` passing
-   `make int_test` passing
-   `make docker_go_lint` passing


cli/bpmetadata/proto/bpmetadata.proto
```diff
@@ -2,87 +2,99 @@ syntax = "proto3";
...
 // - Gen: manually-authored
-// - Gen: partial (contains nested messages that can include both auto-generated and manually authored)
+// - Gen: partial (contains nested messages that can include both auto-generated
+// and manually authored)
 message BlueprintMetadata {
   // APIVersion is the apiVersion field of a metadata file
   // Gen: auto-generated
-  string api_version = 1; // @gotags: json:"apiVersion,omitempty" yaml:"apiVersion,omitempty"
+  string api_version =
+      1;  // @gotags: json:"apiVersion,omitempty" yaml:"apiVersion,omitempty"
```
bharathkkb, line 25: These changes look like formatting changes (maybe IDE?). Could we revert these and follow up in a different PR if it's only cosmetic?

cli/bpmetadata/tfconfig.go
```diff
@@ -188,6 +189,43 @@ func parseBlueprintVersion(versionsFile *hcl.File, diags hcl.Diagnostics) (strin
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, providerData := range hclModule.RequiredProviders {
+		providerVersion := &ProviderVersion{}
+
+		source := providerData.Source
+		if source == "" {
+			fmt.Printf("Not found source in provider settings\n")
```
bharathkkb, line 213: nit: Let's use `Log.Info` instead of `fmt.Printf` here and thoughout

cli/bpmetadata/tfconfig_test.go
```diff
@@ -266,3 +266,36 @@ func TestTFRoles(t *testing.T) {
...
 	}
 }
+
+func TestTFProviderVersions(t *testing.T) {
+	tests := []struct {
+		name                 string
+		configName           string
+		wantProviderVersions []*ProviderVersion
+	}{
+		{
```
bharathkkb, line 276: Could we add a few more cases
- no provider versions at all
- only source and no versions

bharathkkb (approved):
Thanks for the PR @qz267! Few minor comments

cli/bpmetadata/tfconfig_test.go
bharathkkb: Yeah TF would complain and our linter would fail for incomplete/invalid. The valid cases to test here are
- no provider info (TF downloads implied latest according to constraints in child modules if any) - example https://github.com/terraform-google-modules/terraform-google-network/blob/master/codelabs/simple/main.tf
- source is mentioned but no version - example https://github.com/terraform-google-modules/terraform-google-container-vm/blob/b9c7fdd2cd0bd09942440ac372fdee47bde57db9/examples/instance_with_config_file/versions.tf#L19-L21

cli/bpmetadata/tfconfig.go
qz267: done

cli/bpmetadata/tfconfig_test.go
qz267: Tests added and all passing on local runs.

cli/bpmetadata/proto/bpmetadata.proto
qz267: Done

qz267:
@bharathkkb a friendly ping, all the comments addressed, this PR ready for another review.

cli/bpmetadata/tfconfig.go
```diff
@@ -188,6 +189,43 @@ func parseBlueprintVersion(versionsFile *hcl.File, diags hcl.Diagnostics) (strin
 	return "", nil
 }
 
+// parseBlueprintProviderVersions gets the blueprint provider_versions from the provided config
+// from the provider_meta block
```
bharathkkb, line 193:
```suggestion
// from the required_providers block.
```

cli/bpmetadata/tfconfig.go
```diff
@@ -188,6 +189,43 @@ func parseBlueprintVersion(versionsFile *hcl.File, diags hcl.Diagnostics) (strin
...
+func parseBlueprintProviderVersions(versionsFile *hcl.File) ([]*ProviderVersion, error) {
+	var v []*ProviderVersion
+	var diags hcl.Diagnostics
+
+	//parse out the required providers from the config
+	var hclModule tfconfig.Module
+	hclModule.RequiredProviders = make(map[string]*tfconfig.ProviderRequirement)
+	hclModuleDiag := tfconfig.LoadModuleFromFile(versionsFile, &hclModule)
+	diags = append(diags, hclModuleDiag...)
```
bharathkkb, line 202: Is there a need to declare diags seperately?
```suggestion
	// parse out the required providers from the config
	var hclModule tfconfig.Module
	hclModule.RequiredProviders = make(map[string]*tfconfig.ProviderRequirement)
	diags := tfconfig.LoadModuleFromFile(versionsFile, &hclModule)
```

cli/bpmetadata/tfconfig.go
```diff
@@ -188,6 +189,43 @@ func parseBlueprintVersion(versionsFile *hcl.File, diags hcl.Diagnostics) (strin
...
+	err := hasHclErrors(diags)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, providerData := range hclModule.RequiredProviders {
+		providerVersion := &ProviderVersion{}
+
+		source := providerData.Source
```
bharathkkb, line 211: nit: is there a reason to assign to source and then validate. IMO flow could be simplified to

```
if providerData.Source == ""{
   ...
}
if len(providerData. VersionConstraints) == 0{
   ...
}
v = append(v, &ProviderVersion{
source: providerData.Source,
version: strings.Join(providerData.VersionConstraints, ", "),
})
```

cli/bpmetadata/tfconfig.go
```diff
@@ -279,9 +317,22 @@ func getBlueprintRequirements(rolesConfigPath, servicesConfigPath string) (*Blue
 		return nil, err
 	}
 
+	//parse blueprint provider versions
+	versionsFile, diags := p.ParseHCLFile(versionsConfigPath)
```
bharathkkb, line 321: Could we check if `versionsConfigPath` exists and only proceed if versionsConfigPath exists. Ideally I would like to keep role and services even if versions does not exist.

qz267: Done

cli/bpmetadata/tfconfig.go
```diff
@@ -279,10 +310,30 @@ func getBlueprintRequirements(rolesConfigPath, servicesConfigPath string) (*Blue
 		return nil, err
 	}
 
-	return &BlueprintRequirements{
-		Roles:    r,
-		Services: s,
-	}, nil
+	if versionsConfigPath != "" {
```
bharathkkb, line 313: The file path I belive is static and will always be there but we should do something like https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/3081bdc1f6710674330c337aac74484df3e4daa1/cli/util/file.go#L97 to check if its exists.

cli/bpmetadata/tfconfig.go
```diff
@@ -279,10 +310,30 @@ func getBlueprintRequirements(rolesConfigPath, servicesConfigPath string) (*Blue
...
+		if err != nil {
+			return nil, err
+		}
+
+		return &BlueprintRequirements{
+			Roles:            r,
+			Services:         s,
+			ProviderVersions: v,
+		}, nil
+	} else {
```
bharathkkb, line 331: nit: prefer to flip this logic to keep left aligned and can remove else block

```
if file does not exist {
return &BlueprintRequirements{
			Roles:    r,
			Services: s,
		}, nil
}

versionsFile, diags := p.ParseHCLFile(versionsConfigPath)
err = hasHclErrors(diags)
....

```

bharathkkb (approved):
Thanks for updating, just two last comments

cli/bpmetadata/tfconfig.go
qz267: Done

cli/bpmetadata/tfconfig.go
qz267: I applied https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/3081bdc1f6710674330c337aac74484df3e4daa1/cli/bpmetadata/path.go#L23 here, WDYT?

qz267:
hmm, rerun the unit tests on local branch multiple times, all passing, might need to rerun the action job?

apeabody:
This is from the CFT CLI Test:

```
Error! Generated metadata(s) do not match the golden(s).
diff --git a/metadata.yaml b/metadata.yaml
index ffb9656..6059145 100644
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -250,7 +250,7 @@ spec:
       - compute.googleapis.com
       - serviceusage.googleapis.com
     providerVersions:
-      - source: hashicorp/random
-        version: ">= 2.1"
       - source: hashicorp/google
         version: ">= 4.42, < 5.0"
+      - source: hashicorp/random
+        version: ">= 2.1"
```

Looks like maybe update the line ordering in cli/bpmetadata/int-test/goldens/golden-metadata.yaml?

qz267:
Updated golden metadata yml provider data.

bharathkkb:
@qz267 I missed that `RequiredProviders` output was coming from a map. It would be good to stable sort this to prevent any kind of diffs in the future due to ordering.

PR #2502, opened 2024-08-08
feat(tflint_bp_plugin): add min & max parameters

apeabody:


tflint-ruleset-blueprint/rules/terraform_required_version.go
```diff
@@ -48,7 +54,30 @@ const (
...
+	config := &TerraformRequiredVersionConfig{}
+	if err := runner.DecodeRuleConfig(r.Name(), config); err != nil {
+		return err
+	}
+
+	minVersion := minimumTerraformRequiredVersion
+	if config.MinVersion != "" {
+		if _, err := version.NewSemver(config.MinVersion); err != nil {
+			return err
+		} else {
```
bharathkkb, line 66: nit: since there is return above, I think we can omit the else?

```
if _, err := .. {
return err
}
minVersion = config.MinVersion
```

tflint-ruleset-blueprint/rules/terraform_required_version.go
```diff
@@ -48,7 +54,30 @@ const (
...
+
+	maxVersion := maximumTerraformRequiredVersion
+	if config.MaxVersion != "" {
+		if _, err := version.NewSemver(config.MaxVersion); err != nil {
+			return err
+		} else {
+			maxVersion = config.MaxVersion
+		}
+	}
+
```
bharathkkb, line 79: Given we have options for providing min/max version, could we add a log line here with the final selected versions?

bharathkkb (approved):
Thanks @apeabody!

tflint-ruleset-blueprint/rules/terraform_required_version.go
apeabody: Yup, originally I had the order reversed :)

tflint-ruleset-blueprint/rules/terraform_required_version.go
apeabody: Yup, added using the tflint `logger`

PR #2530, opened 2024-08-15
feat: preserve manually added connections fields

tjy9206:
Preserve manually added connections fields in the metadata.yaml files when the metadata files are recompiled for whatever the reason (version updates, dependency changes, etc).

Also added test cases to make sure the logic works in multiple conditions.

### DIFF
--- a/cli/bpmetadata/tfconfig.go
+++ b/cli/bpmetadata/tfconfig.go
@@ -395,3 +395,19 @@ func hasTfconfigErrors(diags tfconfig.Diagnostics) error {
 
 	return nil
 }
+
+// MergeExistingConnections merges existing connections from an old BlueprintInterface into a new one,
+// preserving manually authored connections.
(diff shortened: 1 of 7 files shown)


cli/bpmetadata/tfconfig.go
```diff
@@ -395,3 +395,19 @@ func hasTfconfigErrors(diags tfconfig.Diagnostics) error {
 
 	return nil
 }
+
+// MergeExistingConnections merges existing connections from an old BlueprintInterface into a new one,
+// preserving manually authored connections.
+func MergeExistingConnections(newInterfaces, existingInterfaces *BlueprintInterface) {
```
bharathkkb, line 401: nit: since this is not used outside of this package, lets keep it unexported
```suggestion
func mergeExistingConnections(newInterfaces, existingInterfaces *BlueprintInterface) {
```

bharathkkb (approved):
Thanks for the PR @tjy9206

PR #2591, opened 2024-09-12
feat: sort the variables in the same order from input variables.tf file.

qz267:
Use hclparse to read order of variables from [variables.tf](http://variables.tf/). Sort the metadata.yaml with variables in this obtained order.


q2w:
@qz267 LGTM! Can we add a test scenario for this!

q2w (approved):

cli/bpmetadata/tfconfig.go
```diff
@@ -235,13 +244,38 @@ func getBlueprintInterfaces(configPath string) (*BlueprintInterface, error) {
...
+	for i, block := range variableContent.Blocks {
+		// We only care about variable blocks.
+		if block.Type != "variable" {
+			continue
+		}
+		// We expect a single label which is the variable name.
+		if len(block.Labels) != 1 {
+			Log.Info("Variable block has no name: %v", block)
+		}
+		variableOrderKeys[block.Labels[0]] = i
```
q2w, line 266: Could length of block.Labels be 0? It would be an issue with block.Labels[0] access.

qz267: Above lines checked the `block.Labels` length, let me add `continue` for empty `Lables`.

qz267: Done

cli/bpmetadata/tfconfig.go
```diff
@@ -235,13 +244,38 @@ func getBlueprintInterfaces(configPath string) (*BlueprintInterface, error) {
 		variables = append(variables, v)
 	}
 
+	p := hclparse.NewParser()
```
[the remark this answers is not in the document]
qz267: Done

cli/bpmetadata/tfconfig.go
```diff
@@ -235,13 +244,38 @@ func getBlueprintInterfaces(configPath string) (*BlueprintInterface, error) {
...
+	variableFile, hclDiags := p.ParseHCLFile(filepath.Join(configPath, "variables.tf"))
+	if hclDiags.HasErrors() {
+		Log.Info("Failed to parse HCL: ", "diags: ", hclDiags)
+	}
+	variableContent, _, hclDiags := variableFile.Body.PartialContent(variableSchema)
```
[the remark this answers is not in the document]
qz267: How about calling `log.Error()` instead? Or do you have any suggestions?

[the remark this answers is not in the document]
bharathkkb: My recommendation would be to propagate the error from this function (change signature to (map[string]int, error) and only sort if no error, otherwise log and continue so we degrade gracefully. IRL if this throws an error, we would not reach here as it will likely be caught earlier in tfconfig.LoadModule.

cli/bpmetadata/tfconfig.go
```diff
@@ -256,6 +269,34 @@ func getBlueprintInterfaces(configPath string) (*BlueprintInterface, error) {
...
+		}
+		// We expect a single label which is the variable name.
+		if len(block.Labels) != 1 {
+			Log.Info("Variable block has no name: %v", block)
+			continue
+		}
+		variableOrderKeys[block.Labels[0]] = i
+
+	}
+	Log.Info("Found variables in order: ", "variableOrderKeys: ", variableOrderKeys)
```
bharathkkb, line 296: Let's remove this, I had it just for debugging in the internal poc

cli/bpmetadata/tfconfig_test.go
```diff
@@ -368,3 +368,28 @@ func TestTFIncompleteProviderVersions(t *testing.T) {
...
 	}
 }
+
+func TestTFVariableSortOrder(t *testing.T) {
+	tests := []struct {
+		name         string
+		configPath   string
+		expectOrders map[string]int
+	}{
+		{
```
bharathkkb, line 378: Could we add a few more cases with no outputs and invalid output block declaration?

bharathkkb (approved):
Thanks for the PR! Some minor suggestions

cli/bpmetadata/tfconfig_test.go
qz267: Done

cli/bpmetadata/tfconfig.go
qz267: Done

qz267:
https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/2591#discussion_r1759564728
This is addressed, now the get order function will return `err` to represent any exceptions during the process, and the `getInterfaces` function will gracefully fall back to sort the variables by name if failed to get the original user input orders.

cli/bpmetadata/tfconfig_test.go
```diff
@@ -368,3 +368,48 @@ func TestTFIncompleteProviderVersions(t *testing.T) {
...
+			configPath: "sample-module",
+			expectOrders: map[string]int{
+				"description": 1,
+				"project_id":  0,
+				"regional":    2,
+			},
+			expectError: false,
+		},
+		{
+			name:         "Empty vairable name should create nil order",
```
bharathkkb, line 390:
```suggestion
			name:         "Empty variable name should create nil order",
```

cli/bpmetadata/tfconfig_test.go
```diff
@@ -368,3 +368,48 @@ func TestTFIncompleteProviderVersions(t *testing.T) {
...
+			configPath:   "empty-module",
+			expectOrders: map[string]int{},
+			expectError:  true,
+		},
+		{
+			name:         "No vairable name should create nil order",
```
bharathkkb, line 396:
```suggestion
			name:         "No variable name should create nil order",
```

bharathkkb (approved):

PR #2602, opened 2024-09-18
feat: Add a command to automatically generate type info for output in metadata file

tjy9206:
Add --generate-output-type command to CFT blueprint metadata cli to add output type information to the generated metadata.yaml file.

An example of the auto generated output types can be found in https://github.com/terraform-google-modules/terraform-google-cloud-storage/compare/master...tjy9206:terraform-google-cloud-storage:master.


cli/bpmetadata/cmd.go
```diff
@@ -148,6 +157,21 @@ func generateMetadataForBpPath(bpPath string) error {
...
+		}
+
+		// Update the bpMetadata with the extracted output types
+		for i, output := range bpMetaObj.Spec.Interfaces.Outputs {
+			if outputType, ok := outputTypes[output.Name]; ok {
+				bpMetaObj.Spec.Interfaces.Outputs[i].Type = outputType
+			}
+		}
+	}
+
```
bharathkkb, line 174: Could we keep this in getBlueprintInterfaces which is where the other input/output related logic exists? This will also allow us to test this https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/17331bbd4633f0f56fc07b67700542f8df43521f/cli/bpmetadata/tfconfig_test.go

cli/bpmetadata/parser/state_parser_test.go
```diff
@@ -0,0 +1,180 @@
...
+            {
+              "bar": "string",
+              "id": "string"
+            }
+          ],
+          "number": "number"
+        }
+      ]
+    },
+    "list": {
```
q2w, line 83: nit: let's not use the type as output variable name! 
```suggestion
    "list_output": {
```

cli/bpmetadata/parser/state_parser_test.go
```diff
@@ -0,0 +1,180 @@
...
+      ],
+      "type": [
+        "tuple",
+        [
+          "string",
+          "string"
+        ]
+      ]
+    },
+    "map": {
```
q2w, line 96:
```suggestion
    "map_output": {
```

cli/bpmetadata/parser/state_parser.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		outputTypeMap[name] = pbValue
+	}
+
+	return outputTypeMap, nil
+}
+
+func convertOutputTypeToStructpb(output *tfjson.StateOutput) (*structpb.Value, error) {
+	// Handle nil values explicitly
+	if output.Value == nil {
```
q2w, line 64: Could there be scenarios where Value might be Nil but Type is present?

cli/bpmetadata/parser/state_parser.go
```diff
@@ -0,0 +1,90 @@
...
+		return structpb.NewNullValue(), nil
+	}
+
+	// Handle cases where output.Type is NilType
+	if output.Type == cty.NilType {
```
q2w, line 69: Since we are generating output type for the blueprint, having NilType for an output variable doesn't provide any value? Why do we have a output variable with NilType?

cli/bpmetadata/parser/state_parser_test.go
```diff
@@ -0,0 +1,180 @@
...
+        }
+      ]
+    },
+    "list": {
+      "value": [
+        "foo",
+        "bar"
+      ],
+      "type": [
+        "tuple",
```
q2w, line 89: Since we use list type for input variables, can we have same for output variables too?

cli/bpmetadata/parser/state_parser.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Remove the top-level "outputs" field
+	delete(rawState, "outputs")
+
+	// Now marshal the updated map back to JSON
+	updatedStateData, err := json.MarshalIndent(rawState, "", "  ")
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal updated state data: %w", err)
+	}
+
```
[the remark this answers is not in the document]
tjy9206: There are two major reasons for the preprocessing:
1. if `format_version` key is not in the generated tfstate the marshal will fail;
2. the generated tfstate has this `outputs` field, but in tfjson.State struct it has an extra layer around outputs called "values": https://github.com/hashicorp/terraform-json/blob/main/state.go#L37

These were actually the major blockers while I was implementing the feature, thus we need this preprocessing. I can extract the functionality to a preprocessing function and add comments for the rationale if you think that is necessary.

cli/bpmetadata/cmd.go
```diff
@@ -180,6 +204,43 @@ func generateMetadataForBpPath(bpPath string) error {
...
 }
 
+// Extract output types from terraform.tfstate
+func extractOutputTypesFromState(bpPath string) (map[string]*structpb.Value, error) {
+	outputTypes := make(map[string]*structpb.Value)
+
+	// Construct the path to the test/setup directory
+	tfDir := filepath.Join(bpPath)
+
+	// testing.T chekcs verbose flag to determine it's mode. Add this line as a flags initializer
```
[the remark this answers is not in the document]
tjy9206: Good catch!

cli/bpmetadata/parser/state_parser.go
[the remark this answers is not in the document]
bharathkkb: Hmm does that mismatch still exist if you do a `terraform show -json` and use the json output from there rather than reading the statefile from disk?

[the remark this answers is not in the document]
tjy9206: Yes the result from `terraform show -json` after running `terraform apply` does have the `outputs` wrapped in `values`.

Do you have a recommended way to do this in tft? Do I need to create a function similar to `tft.Test()` and wrap that command in as well?

cli/bpmetadata/cmd.go
tjy9206: Make sense to me. Made the changes and added tests in [this commit](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/2602/commits/7dbdcb144b42d9eaae23296247605261a1850e07)

cli/bpmetadata/parser/state_parser_test.go
tjy9206: Good catch!

cli/bpmetadata/parser/state_parser.go
tjy9206: I don't think this is a valid case since `tfjson.StateOutput` only have three fields - `Sensitive`, `Value` and `Type`, thus if `Value` is `nil` then the output should not exist. @bharathkkb to confirm.

cli/bpmetadata/parser/state_parser.go
tjy9206: I don't think this is a valid case either. Just added this for robustness of the code. If this is indeed not a case I can either remove this or not set the type in `bpInterface`. @bharathkkb WDYT?

cli/bpmetadata/parser/state_parser_test.go
tjy9206: Yeah it's weird but it's probably just how Terraform represents list in `.tfstate` files. For example the `urls_list` in GCS config is represented as below:

```
    "urls_list": {
      "value": [
        "gs://jieyutian-test-bucket-7"
      ],
      "type": [
        "tuple",
        [
          "string"
        ]
      ]
    }
```

Does it make sense to you?

q2w: @tjy9206 Ohkay! I am wondering if can we transform `tuple` to `list` ourselves before writing metadata. Terraform might be representing list as tuple in .tfstate because the size of list output would be fixed as the deployment has already happened. But we are adding type to metadata and before deployment there is no way of definitely knowing the size of the list output. @bharathkkb WDYT?

tjy9206: It will be tricky though since we probably don't know ahead of time whether it should be a tuple of, say, two strings (key-value pair for example) or a list of strings. I would propose we keep the output generated by Terraform and have a separate post-processing. WDYT?

cli/bpmetadata/parser/state_parser.go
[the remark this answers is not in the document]
bharathkkb: @tjy9206 You can use [1] to grab output of show. Something like `terraform.Show(t, root.GetTFOptions())` should work

[1] https://pkg.go.dev/github.com/gruntwork-io/terratest@v0.47.1/modules/terraform#Show

cli/bpmetadata/parser/state_parser.go
bharathkkb: Yeah I believe something like
```
output "test1" {
  value = null
}
``` 
will cause value to be omitted entirely rather than having null in output.

cli/bpmetadata/parser/state_parser.go
bharathkkb: This is also not a valid case AFAIK but we can keep it here as a check.

cli/bpmetadata/parser/state_parser_test.go
bharathkkb: I discussed this with @q2w today and we agreed to keep as is for now (without any post processing)

bharathkkb:
@tjy9206 Overall LGTM, just that one request to switch to using TF show so we can avoid the preprocessing workaround.

cli/bpmetadata/parser/state_parser.go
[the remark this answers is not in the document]
tjy9206: Thanks for the reference! Modified accordingly.

bharathkkb (approved):

PR #2610, opened 2024-09-20
feat: adds support to multi-level submodule

amandakarina:
Adds support to sub-modules in structure:
terraform-google-modules/example/step/module/bar-module


apeabody:
@bharathkkb FYI: This is for compatibility with repo like https://github.com/GoogleCloudPlatform/terraform-google-enterprise-application

infra/module-swapper/cmd/testdata/example-with-steps-submodules/examples/example-one/main.tf
```diff
@@ -0,0 +1,27 @@
+module "test-module" {
+  source  = "terraform-google-modules/example-with-steps-submodules/google"
+  version = "~> 3.2.0"
+
+  project_id   = var.project_id # Replace this with your project ID in quotes
```
renato-rudnicki, line 5: Based on the comment `# Replace this with your project ID in quotes` I'm wondering if the best approach would be to hardcode the _project_id_ on the main.tf or keep the it as it is (variables) and add the value into a `terraform.tfvars` file.

renato-rudnicki (commented):
Only a question, but the rest looks good.

infra/module-swapper/cmd/testdata/example-with-steps-submodules/examples/example-one/main.tf
apeabody: Thanks @renato-rudnicki - This is testdata, and this particular field isn't actually critical to the test.

apeabody:
@amandakarina - I believe this can be closed?

PR #2613, opened 2024-09-23
feat: preserve existing alt defaults in display yaml.

qz267:
Feature requested in http://b/368374398


cli/bpmetadata/display.go
```diff
@@ -34,3 +34,24 @@ func createTitleFromName(name string) string {
...
+	for i, variable := range newInput.Variables {
+		for _, existingVariable := range existingInput.Variables {
+			if variable.Name == existingVariable.Name && existingVariable.AltDefaults != nil {
+				newInput.Variables[i].AltDefaults = existingVariable.AltDefaults
+			}
+		}
+	}
+}
+
+func mergeExistingDisplayData(newInput, existingInput *BlueprintUIInput) {
```
bharathkkb, line 54: nit: I would recommend removing this for now and adding later if we do end up needing other merge functions.

bharathkkb (approved):

cli/bpmetadata/display.go
qz267: Done

PR #2617, opened 2024-09-25
refactor: use `-chdir` in terraform_validate wrapper

wyardley:
- Use Terraform's `-chdir` option rather than `cd`


infra/build/developer-tools/build/scripts/task_wrapper_scripts/terraform_validate
```diff
@@ -19,8 +19,5 @@
 # files to validate.  This directory is assumed to be a root module.
 
 set -eu
-curdir=$(pwd)
-cd "${1}"
-flock -x /workspace/.terraform.lock -c "terraform init -backend=false >/dev/null"
-flock -s /workspace/.terraform.lock -c "terraform validate"
-cd "$curdir"
+flock -x /workspace/.terraform.lock -c "terraform init -compact-warnings -upgrade -chdir=$1 -backend=false >/dev/null"
```
wyardley, line 22: note: these flags could be also set via `TF_CLI_ARGS_init` env var.

apeabody: Thanks @wyardley - Using `-chdir` make sense.  Let me ask around about `-upgrade` as that would be a change in  behavior that might not be desired for CI linting (although I can understand for local use).

wyardley: I had that thought, but since the lockfiles are .gitignored, I don’t think it should have any practical effect in terms of changing behavior?

I thought about removing the wrapper entirely, but you’d added in those flock calls since the version of the container I had locally, so I guess makes sense to keep it?

Creating the lockfiles locally also creates some issues more generally, but I don’t see any easy way to fix it other than removing them explicitly, which seems less than ideal.

wyardley:
Docs generation is another place where I _think_ the pushd / popd might not be needed and where the code might be simpler / more readable without it.

infra/build/developer-tools/build/scripts/task_wrapper_scripts/terraform_validate
apeabody: In a few repos (such as this one), we "force" add lock files. E.g. https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/terraform/test-org/org/.terraform.lock.hcl

Yeah, flock is needed as the terraform provider cache is not concurrent safe (https://developer.hashicorp.com/terraform/cli/config/config-file#provider-plugin-cache).  In theory those could be made contingent on the presence of the `ENABLE_PARALLEL=1` env.

wyardley: > In a few repos (such as this one), we "force" add lock files

Got it. yes, that could cause some inconsistent behavior in those cases then 😞 
An overengineered solution could be to check if the lockfile comes from git or not, and add `-upgrade` to the arguments if it doesn't, but that seems kind of messy.

> as the terraform provider cache is not concurrent safe
Hah yes, something I've [come across](https://github.com/hashicorp/terraform/issues/25849) as well... given the small number of different providers, maybe the `terraform providers mirror` workaround mentioned there would work? Sharing the cache across CI runs could also speed things up.

In theory, once the cache is populated once with all the necessary providers, there shouldn't be any other issues with parallel init.

I have noticed that locally, at least (maybe because of Docker's filesystem) this stuff is excruciatingly slow, which is one reason I've poked around at some of this tooling.

I get the idea to keep things standard across projects, minimizing local setup, and using the custom tflint plugin, but if some of the lint checks could be run locally instead, my suspicion is that they'd be much faster.

apeabody: Perhaps, although I suspect major component is the time for `init` checking the registry, etc.

It should be possible to test the theory by running the same scripts (https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/tree/master/infra/build/developer-tools/build/scripts) outside of docker and comparing the time.

wyardley: maybe the `-upgrade` thing could be controlled by a flag?

I could revert that part of the change for now if you think it's too risky, and can revisit it if / when there's a better idea of how to proceed?

apeabody: Yeah - If you want to move `-upgrade` to a separate PR, I can get the rest of the merged now.  I was considering an ENV as well, so it could be selectively enabled/disabled.

wyardley: At that point, don't even need it in the script, could maybe even just set `TF_CLI_ARGS_init="upgrade"` if it doesn't interfere with anything else.

infra/build/developer-tools/build/scripts/task_wrapper_scripts/terraform_validate
```diff
@@ -19,8 +19,5 @@
 # files to validate.  This directory is assumed to be a root module.
 
 set -eu
```
wyardley, line 21: @apeabody do you think setting `o pipefail` is a good idea here too?

infra/build/developer-tools/build/scripts/task_wrapper_scripts/terraform_validate
wyardley: For now, I've just created an issue so it can be tracked.

infra/build/developer-tools/build/scripts/task_wrapper_scripts/terraform_validate
apeabody: Hmm, this can also be called via parallel (used by https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/Makefile#L72), so we should just be careful that use case doesn't cause a probleme: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/build/scripts/task_helper_functions.sh#L194

wyardley:
updated the commit type to `refactor`, since it shouldn't change the behavior (ideally)

infra/build/developer-tools/build/scripts/task_wrapper_scripts/terraform_validate
wyardley: since there are no pipes, it probably doesn't matter too much? I usually try to use it as a matter of habit, but probably can leave it out here.

apeabody: Yeah, it should be fine.  If you have done a tested against a parallel linted repo, that's good enough for me.

apeabody (changes requested):
Thanks for the contribution @wyardley!

Just to be safe/proper, let's bump the patch here: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/Makefile#L59, and then this is ready to merge.

wyardley:
Ok, and then bump the version in #2619 again after merge? Right now, they would both be at the same patch level.

apeabody:
> Ok, and then bump the version in #2619 again after merge? Right now, they would both be at the same patch level.

Exactly, a bit of extra work, but I'd like to get this PR merged/released today.

apeabody (approved):

apeabody (approved):
Thanks for the contribution @wyardley!

PR #2626, opened 2024-09-26
docs(blueprint-test): update notes in README

wyardley:
This is trivial, but I think makes the notes a bit more readable?

Can preview via https://github.com/wyardley/cloud-foundation-toolkit/tree/wyardley/docs/test_readme/infra/blueprint-test#readme


apeabody:
/gcbrun

(we should update this trigger to ignore markdown files)

PR #2637, opened 2024-10-07
fix: Version config path should look for versions.tf in current blueprint dirctory

q2w:
In blueprint metadata generation step, versions.tf needs to be loaded for generating version metadata. 
Currently versions.tf file is being only searched in the root directory of the blueprint. This fails to generated version metadata for the submodules.
This feature was added in https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/2496


q2w:
@qz267 PTAL!

qz267 (approved):
Thanks @q2w for the fix!

bharathkkb (approved):

qz267:
@tjy9206 do we need to rerun the generate output type after ^^?

tjy9206:
> @tjy9206 do we need to rerun the generate output type after ^^?

I don't think so since the generated output types should have nothing to do with the version, but we should double check to confirm. Will try generating again when I get a chance to setup the overheads.

PR #2647, opened 2024-10-09
feat: Preserve output types in bpmetadata interface

tjy9206:
Preserve output type when generating metadata without --generate-output-type flag

### DIFF
--- a/cli/bpmetadata/tfconfig.go
+++ b/cli/bpmetadata/tfconfig.go
@@ -522,6 +522,25 @@ func mergeExistingConnections(newInterfaces, existingInterfaces *BlueprintInterf
 	}
 }
 
+// mergeExistingOutputTypes merges existing output types from an old BlueprintInterface into a new one,
+// preserving manually authored types.
+func mergeExistingOutputTypes(newInterfaces, existingInterfaces *BlueprintInterface) {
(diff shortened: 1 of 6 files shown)


bharathkkb (approved):
Thanks for catching this!

cli/bpmetadata/tfconfig.go
```diff
@@ -522,6 +522,25 @@ func mergeExistingConnections(newInterfaces, existingInterfaces *BlueprintInterf
...
+	}
+
+	existingOutputs := make(map[string]*BlueprintOutput)
+	for _, output := range existingInterfaces.Outputs {
+		existingOutputs[output.Name] = output
+	}
+
+	for i, output := range newInterfaces.Outputs {
+		if existingOutput, ok := existingOutputs[output.Name]; ok && existingOutput.Type != nil {
+			newInterfaces.Outputs[i].Type = existingOutput.Type
```
q2w, line 539: This will overwrite type info if the output-type was generated.. In effect this keeps the existing type info  even if the output type has changed.
Let's add a condition to only update type info when it is not present in the newInterfaces. Let's also add a test case for this.

tjy9206: Good catch! Modified accordingly.

PR #2661, opened 2024-10-18
feat: Support display metadata AlternateDefault validate

q2w:
This is a follow-up PR of https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/2453. The AlternateDefault value field is of type `google.protobuf.Value`. Similar to other fields of same type, it requires workaround so that JsonSchema doesn't treat it as object.


apeabody (approved):
Thanks @q2w!

PR #2669, opened 2024-10-24
feat: Bypass `cft test lint` rules

qz267:
Add `BLUEPRINT_LINT_DISABLE` env var to bypass `cft test lint` rules when set to 1.

### DIFF
--- a/cli/bptest/lint_interface.go
+++ b/cli/bptest/lint_interface.go
@@ -1,7 +1,9 @@
 package bptest
 
 import (
+	"fmt"
 	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/cli/bpmetadata"
+	"os"
 )
 
 // lintRule defines the common interface for all metadata lint rules.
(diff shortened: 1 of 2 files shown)


cli/bptest/lint_interface_test.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/cli/bpmetadata"
+	"github.com/stretchr/testify/assert"
+)
+
+type MockLintRule struct {
```
bharathkkb, line 12:
```suggestion
type mockLintRule struct {
```

cli/bptest/lint_interface.go
```diff
@@ -30,6 +32,11 @@ func (r *lintRunner) RegisterRule(rule lintRule) {
 // Run runs all the registered rules on the provided context.
 func (r *lintRunner) Run(ctx lintContext) []error {
 	var errs []error
+	if os.Getenv("BLUEPRINT_LINT_DISABLE") == "1" {
```
bharathkkb, line 35: nit: Let's move this to a const

bharathkkb (approved):
Thanks @qz267!

cli/bptest/lint_interface_test.go
qz267: Done

cli/bptest/lint_interface.go
qz267: Done

PR #2689, opened 2024-11-11
feat(blueprint-test): add cai helper

apeabody:


infra/blueprint-test/pkg/cai/cai.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+	"time"
+
+	"github.com/GoogleCloudPlatform/cloud-foundation-toolkit/infra/blueprint-test/pkg/gcloud"
+	"github.com/tidwall/gjson"
+)
+
+type CmdCfg struct {
+	sleep     int      // minutes to sleep prior to CAI retreval. default: 2
+	assetType string   // asset type to retrieve. default: all
```
bharathkkb, line 32: Not super familiar with CAI but would this allow to retrieve two asset types or should it be slice of strings?

bharathkkb (approved):

infra/blueprint-test/pkg/cai/cai.go
apeabody: Good idea - I'll also need to re-work the newCmdConfig, but the underlying ` --asset-types` can accept a comma delimitated list.

PR #2706, opened 2024-11-19
feat(blueprint-test): add JSONPathEqs()

apeabody:


infra/blueprint-test/pkg/golden/golden.go
```diff
@@ -156,3 +157,16 @@ func (g *GoldenFile) JSONEq(a *assert.Assertions, got gjson.Result, jsonPath str
 	gfData := gf.Get(jsonPath).String()
 	a.Equal(gfData, gotData, fmt.Sprintf("expected %s to match fixture %s", jsonPath, gfData))
 }
+
+// JSONEqs asserts that json content in jsonPaths for got and goldenfile are the same
+func (g *GoldenFile) JSONEqs(a *assert.Assertions, got gjson.Result, jsonPaths []string) {
```
bharathkkb, line 162: nit: Maybe JSONPathEqs might be a better name?

infra/blueprint-test/pkg/golden/golden_test.go
```diff
@@ -133,3 +133,30 @@ func TestJSONEq(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := assert.New(t)
+			os.Setenv(gfUpdateEnvVar, "true")
+			defer os.Unsetenv(gfUpdateEnvVar)
+			got := NewOrUpdate(t, tt.data, tt.opts...)
+			defer os.Remove(got.GetName())
+			got.JSONEqs(assert, utils.ParseJSONResult(t, tt.want), tt.eqPaths)
```
bharathkkb, line 159: Not sure I fully grokked this- is want the newer data version we are testing against? Maybe we should rename the field name?

Also should we add a test case where it does not match with golden?

bharathkkb (approved):

infra/blueprint-test/pkg/golden/golden.go
apeabody: SG

infra/blueprint-test/pkg/golden/golden_test.go
apeabody: I revised the test logic, let me know if it makes more sense now.

Working on how-to include a negative test case as the `JSONPathEqs()` includes an `assert`.

PR #2714, opened 2024-11-25
feat: Add roles requirement in blueprint metadata in a order

q2w:


cli/bpmetadata/tfconfig.go
```diff
@@ -441,9 +441,36 @@ func parseBlueprintRoles(rolesFile *hcl.File) ([]*BlueprintRoles, error) {
...
+		// 1. Sort by Level
+		if r[i].Level < r[j].Level {
+			return true
+		}
+		if r[i].Level > r[j].Level {
+			return false
+		}
+
+		// 2. If Levels are equal, sort by the first role (if available)
+		if len(r[i].Roles) > 0 && len(r[j].Roles) > 0 {
```
bharathkkb, line 460: Since we just want them to be stable sorted, would it be better to sort by len of slice instead to simplify?

cli/bpmetadata/tfconfig.go
```diff
@@ -441,9 +441,36 @@ func parseBlueprintRoles(rolesFile *hcl.File) ([]*BlueprintRoles, error) {
 		break
 	}
 
+	sortBlueprintRoles(r)
 	return r, nil
 }
 
+// Sort blueprint roles.
+func sortBlueprintRoles(r []*BlueprintRoles) {
+	sort.SliceStable(r, func(i, j int) bool {
```
bharathkkb, line 450: Let's add a unit test for this exercising all conditions so incase something changes in the future we can avoid spurious diffs across our repos.

bharathkkb (approved):

cli/bpmetadata/tfconfig.go
q2w: The probability of len being same is higher. Thus it will risky to just sort on len. IMO sorting on first role would result in better stability. The chance of having same role in 2 different level would be very minimal.

q2w: Have updated the logic for sorting.

* first sort based on level.. Eg. `folder` comes before `project`
* if level is same, sort on the length of the roles.
* if length of the roles are same, sort on the first role if it is available

IMO this would result in mostly stable sorting. Let me know what you think!

cli/bpmetadata/tfconfig.go
q2w: Added UTs, PTAL!

bharathkkb (approved):

PR #2715, opened 2024-11-25
chore: Exclude metadata.display.yaml from GCB trigger

q2w:


bharathkkb (approved):
This has been applied

PR #2748, opened 2024-12-11
fix(gcloud): support Run cmd with quotes

apeabody:


infra/blueprint-test/pkg/gcloud/gcloud.go
```diff
@@ -93,7 +94,10 @@ func RunCmdE(t testing.TB, cmd string, opts ...cmdOption) (string, error) {
 		t.Fatal(err)
 	}
 	// split command into args
-	args := strings.Fields(cmd)
+	args, err := shellwords.Parse(cmd)
```
bharathkkb, line 97: Docs seem to suggest env parsing is opt-in but wanted to confirm

bharathkkb (approved):

infra/blueprint-test/pkg/gcloud/gcloud.go
apeabody: Yeah, for ENV parsing, my understanding (also looking at the source) is to instead use `shellwords.ParseWithEnvs()`.

PR #2757, opened 2024-12-16
feat: update toggleUsingVariable to be an array

vicentefb:
In order to support multiple toggles let's define an array of toggles


vicentefb:
@g-awmalik

g-awmalik (approved):

PR #2816, opened 2025-01-22
chore(CI): switch to default codeql

apeabody:


infra/utils/fbf/go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/GoogleCloudPlatform/cloud-foundation-toolkit/infra/utils/fbf
 
 go 1.22
```
g-awmalik, line 3: Do we need this if you're containing the toolchain to 1.23.4?

g-awmalik (approved):

infra/utils/fbf/go.mod
apeabody: Yeah - The toolchain is a recent requirement, it indicates what version of the GO binary last modified the file, not which version of GO is required.

PR #3012, opened 2025-05-22
feat(test-org): refactor to support additional global admin group(s)

apeabody:


infra/terraform/modules/repositories/main.tf
```diff
@@ -106,6 +115,23 @@ resource "github_repository_collaborator" "maintainers" {
...
+  for_each = {
+    for v in local.admin_groups : "${v.repo}/${v.admin_group}" => v
+  }
+  repository = each.value.repo
+  team_id    = each.value.admin_group
+  permission = "maintain"
+}
+
+# manually applied to repo
+# resource "github_team_repository" "super_admin" {
```
bharathkkb, line 128: Do we need to keep this?

bharathkkb (approved):
Thanks @apeabody!

infra/terraform/modules/repositories/main.tf
apeabody: I can drop the commented out code.

PR #3041, opened 2025-06-04
feat: per module requirements config

qz267:
This CL implement design in [go/per-module-requirement](https://goto.google.com/per-module-requirement).

Changes including:
- introduced `locals.per_module_roles` to `test/setup/iam.tf`and `locals.per_module_services` to `test/setup/main.tf` to carry the per module requirements configs
- update blueprint parser to parse the per module config
- added related tests.


cli/bpmetadata/tfconfig.go
```diff
@@ -620,3 +665,48 @@ func generateTFState(bpPath string) ([]byte, error) {
 
 	return stateData, nil
 }
+
+// Helper function to extract the local per module configs
+func extractModuleLocalList(file *hcl.File, localKey string, moduleName string) ([]string, error) {
```
q2w, line 669: When will this method be used?

cli/bpmetadata/tfconfig_test.go
```diff
@@ -408,6 +409,107 @@ func TestSortBlueprintRoles(t *testing.T) {
 	}
 }
 
+func TestParseBlueprintRoles_PerModuleMode_HappyPath(t *testing.T) {
+	const hclContent = `
+locals {
+  per_module_roles = {
```
q2w, line 415: Where would test only roles be defined?

cli/bpmetadata/tfconfig.go
```diff
@@ -415,28 +428,60 @@ func parseBlueprintRoles(rolesFile *hcl.File) ([]*BlueprintRoles, error) {
 			return nil, err
 		}
 
-		for k := range iamAttrs {
-			var iamRoles []string
-			attrValue, _ := iamAttrs[k].Expr.Value(nil)
-			if !attrValue.Type().IsTupleType() {
-				continue
+		// Special branch for perModuleMode
+		if perModuleMode {
```
q2w: What would happen if someone runs in perModuleMode on an existing terraform blueprint which has roles in older format.

cli/bpmetadata/tfconfig.go
qz267: This is a helper to parse the per module local configs, will be called by parseBlueprintRoles and parseBlueprintServices.

cli/bpmetadata/tfconfig_test.go
qz267: `test_only` roles and services are not necessary in this case, currently the `int_test_roles` carries all the test required roles. We might wanna flatten the per module roles and concat them with the existing int_test_roles if needed.

In short, this per module roles not affecting the integration test roles configs, it is helping CFT to generate bpmetadata.

cli/bpmetadata/tfconfig.go
qz267: there should be some quick checks, for example the `per_module_roles` local need to be defined. I'll update accodingly.

cli/bpmetadata/tfconfig_test.go
q2w: I am not able to place where would int_test_roles be present. Can you update this locals block with everything including roles needed by tests?

cli/bpmetadata/tfconfig.go
q2w: Will there be a followup PR for this?

qz267: right, moved out from this PR.

cli/bpmetadata/tfconfig_test.go
qz267: Not sure why that matters. 

Updated the test case using a full iam.tf content.

cli/bpmetadata/tfconfig.go
```diff
@@ -396,51 +402,64 @@ func getBlueprintRequirements(rolesConfigPath, servicesConfigPath, versionsConfi
...
 		if err != nil {
 			return nil, err
 		}
 
-		for k := range iamAttrs {
+		seen := map[string]struct{}{}
+		for _, role := range append(rootRoles, moduleRoles...) {
+			seen[role] = struct{}{}
+		}
+		var combined []string
```
q2w, line 416: What is the reason for combining this? Root module is an independent module. A module is not dependent on the root module.

cli/bpmetadata/tfconfig_test.go
q2w: Just wanted to understand your thoughts. I see you have put the module related roles in `int_required_roles` too. I believe we discussed about it offline that you would be removing this from `int_required_roles`.

cli/bpmetadata/tfconfig.go
qz267: http://screen/8yxur4cexeszGL5 is the design detail of the submodule roles, the implementation aligns well here.

Are you saying the sub module roles should not merge the root roles? Could you help clarify?

cli/bpmetadata/tfconfig_test.go
qz267: I think we agreed on that keep all the sub module roles and flatten the per module role map then merge to `int_required_roles` for testing purpose. Please check https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/3041/files#diff-49f29e3851f3242ff7b138f85242b64698ec38b39a9839137f6e26b983026b88R428-R436 these lines, it fully aligned with above.

Or feel free suggest changes you expected.

cli/bpmetadata/tfconfig.go
q2w: Yes, submodules are independent of root module. You don't need to combine roles with root module. 

Take an example for Memorystore, while the root module creates redis the memcache sumodule creates memcache which is doesn't depend at all on the root module. This is the case for every terraform blueprint.

cli/bpmetadata/tfconfig_test.go
q2w: Thanks! Makes sense!

cli/bpmetadata/tfconfig.go
```diff
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"sort"
+	//"slices"
```
bharathkkb: nit: remove commented code

cli/bpmetadata/tfconfig.go
```diff
@@ -396,51 +402,60 @@ func getBlueprintRequirements(rolesConfigPath, servicesConfigPath, versionsConfi
 }
 
 // parseBlueprintRoles gets the roles required for the blueprint to be provisioned
-func parseBlueprintRoles(rolesFile *hcl.File) ([]*BlueprintRoles, error) {
+func parseBlueprintRoles(rolesFile *hcl.File, perModuleMode bool, moduleName string) ([]*BlueprintRoles, error) {
```
bharathkkb, line 404: Since there is little to no shared code, I would move this to a separate function (parsePerModuleBlueprintRoles) rather than have a boolean that has logic in here and early returns.

cli/bpmetadata/tfconfig.go
```diff
@@ -620,3 +652,47 @@ func generateTFState(bpPath string) ([]byte, error) {
...
+	if diags.HasErrors() {
+		return nil, diags
+	}
+
+	for _, block := range content.Blocks {
+		attrs, diags := block.Body.JustAttributes()
+		if diags.HasErrors() {
+			return nil, diags
+		}
+		if attr, ok := attrs[localKey]; ok {
```
bharathkkb: nit: flip this conditional with early return for less nesting
```
if !ok{
continue
}
...
```

cli/bpmetadata/tfconfig.go
```diff
@@ -620,3 +652,47 @@ func generateTFState(bpPath string) ([]byte, error) {
...
+			val, _ := attr.Expr.Value(nil)
```
bharathkkb: Doesn't attr.Expr.Value return diags? If so we should check it since otherwise val could be nil

cli/bpmetadata/tfconfig.go
```diff
@@ -620,3 +652,47 @@ func generateTFState(bpPath string) ([]byte, error) {
...
+			if val.Type().IsObjectType() {
```
bharathkkb: Flip conditionals here and throughout to reduce logic nesting.

cli/bpmetadata/tfconfig_test.go
```diff
@@ -408,6 +409,283 @@ func TestSortBlueprintRoles(t *testing.T) {
...
+	roles, err := extractModuleLocalList(file, "per_module_roles", "run")
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	expected := []string{"roles/run.invoker", "roles/logging.logWriter"}
+	if len(roles) != len(expected) {
+		t.Errorf("got %d roles, want %d", len(roles), len(expected))
+	}
+	for _, want := range expected {
```
bharathkkb: Could we use https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch?

cli/bpmetadata/tfconfig.go
qz267: done

cli/bpmetadata/tfconfig.go
qz267: Agree there are not many shared code in this function, the reason I have both `parseBlueprintRoles` and `parseBlueprintServices` is trying to make the `getBlueprintRequirements` more readable, other wise there will be multiple blocks of HCL file parsing code in the `getBlueprintRequirements` implementation, WDYT?

cli/bpmetadata/tfconfig.go
qz267: done

cli/bpmetadata/tfconfig.go
qz267: done

cli/bpmetadata/tfconfig.go
qz267: done

cli/bpmetadata/tfconfig_test.go
qz267: done

cli/bpmetadata/tfconfig.go
bharathkkb: My suggestion was actually just for the code that is activated by perModuleMode boolean. That does not seem to parse hcl and returns early. 

```
if perModuleMode {
		moduleRoles, err := extractModuleLocalList(rolesFile, "per_module_roles", moduleName)
		...
		return r, nil
}
```

I was thinking instead of passing in the perModuleMode to this function to activate this code path, we could have a different function like `parsePerModuleBlueprintRoles` that contains this logic. Very minor nit, so feel free to ignore.

bharathkkb (approved):

PR #3103, opened 2025-07-17
fix: lint

apeabody:


infra/blueprint-test/pkg/tft/terraform_test.go
```diff
@@ -187,9 +191,17 @@ func TestSetupOverrideString(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			emptyDir := newTestDir(t, "empty*", "")
+			t.Cleanup(func() {
+				if err := os.RemoveAll(emptyDir); err != nil {
+					t.Logf("Error removing temp dir %s: %v", emptyDir, err)
+				}
+			})
 			setupDir := newTestDir(t, "setup-*", tt.tfOutputs)
-			defer os.RemoveAll(emptyDir)
```
g-awmalik, line 191: Are you missing the removal of `emptyDir`?

infra/blueprint-test/pkg/tft/terraform_test.go
```diff
@@ -242,9 +254,17 @@ func TestSetupOverrideList(t *testing.T) {
...
```
g-awmalik, line 246: same as above

g-awmalik (approved):

infra/blueprint-test/pkg/tft/terraform_test.go
apeabody: It's moved to the following `if` statement: `if err := os.RemoveAll(setupDir); err != nil {`.  That way we can check if it an error is thrown.

infra/blueprint-test/pkg/tft/terraform_test.go
apeabody: It's moved to the following `if` statement: `if err := os.RemoveAll(setupDir); err != nil {`

PR #3109, opened 2025-07-23
chore: add adc admin group

bharathkkb:
Adds adc admin groups for tgm and gcp orgs

### DIFF
--- a/infra/terraform/test-org/github/protection.tf
+++ b/infra/terraform/test-org/github/protection.tf
@@ -15,8 +15,8 @@
  */
 
 locals {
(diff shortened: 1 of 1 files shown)


apeabody (approved):

bharathkkb:
@apeabody I got some errors using node_id so switched to id based on example in https://registry.terraform.io/providers/integrations/github/latest/docs/resources/team_repository

bharathkkb:
This has been applied

apeabody:
> This has been applied

Interesting, will have to look into the difference between `node_id` and `id`.

PR #3146, opened 2025-08-22
feat: Add UI metadata extensions for GKE Namespace, StorageClass and Servic…

skywardpixel:
…eAccount.


cli/bpmetadata/proto/bpmetadata_ui_ext.proto
```diff
@@ -41,6 +41,9 @@ enum ExtensionType {
 
   // GKE related.
   ET_GKE_CLUSTER = 20;
+  ET_GKE_NAMESPACE = 23;
```
g-awmalik, line 44: Is there a reason we're skipping 21, 22?

cli/bpmetadata/proto/bpmetadata_ui_ext.proto
```diff
@@ -76,6 +79,9 @@ message GooglePropertyExtension {
   GCEDiskTypeExtension       gce_disk_type = 17; // @gotags: json:"gceDiskType,omitempty" yaml:"gceDiskType,omitempty"
   GCELocationExtension       gce_location = 18; // @gotags: json:"gceLocation,omitempty" yaml:"gceLocation,omitempty"
   GKEClusterExtension        gke_cluster = 19; // @gotags: json:"gkeCluster,omitempty" yaml:"gkeCluster,omitempty"
+  GkeNamespaceExtension      gke_namespace = 20; // @gotags: json:"gkeNamespace,omitempty" yaml:"gkeNamespace,omitempty"
```
g-awmalik, line 82: These seem off by 1. Please take another look if this should be addressed.

g-awmalik (approved):
LGTM but please take a look at the numbering for the attributes.

cli/bpmetadata/proto/bpmetadata_ui_ext.proto
skywardpixel: My understanding is not all extension types have additional metadata (for example, ET_EMAIL_ADDRESS), so these field numbers don't match the enum option numbers. Here (and above) I'm simply using the next 3 numbers.

cli/bpmetadata/proto/bpmetadata_ui_ext.proto
skywardpixel: 21 and 22 are already used above (ET_CREATE_RESOURCE, ET_GCE_LOCATION).

g-awmalik (approved):

PR #3194, opened 2025-09-17
feat: sort roles and services in metadata.yamls

martijnvans:
The prior implementation of parseBlueprintRoles was outputting the roles in golang map iteration order, which is deliberatey randomized in practice. Sorting the roles will mean that they are not output in a scrambled order each time the metadata is regenerated.

parseBlueprintServices was already sorting the output services.


apeabody (approved):
Thanks @martijnvans! - I'd noticed they were non-idempotent

q2w (approved):

PR #3195, opened 2025-09-17
chore: Update Cloud Run maintainers

trshafer:
This should include the new list of maintainers. If there is a way to add a group that would be preferred.


apeabody:
applied

apeabody (approved):
Thanks @trshafer!

PR #3201, opened 2025-10-07
feat: Support for dispatching to isolated projects/service accounts

martijnvans:
This PR adds support for a new testing mode, in which test setup creates multiple projects and service accounts, and each module's tests run in these projects, isolated from one another. This is to allow the projects to have separate services enabled, and to allow the service accounts to have different permissions enabled.

An example of how this would be used in a specific repo is here: https://github.com/GoogleCloudPlatform/terraform-google-cloud-run/pull/408/files.

Unfortunately, I didn't see a simpler way to acoomplish the task of "searching for all modules currently being tested" than doing a graph search over modules. This diagram illustrates the possibilities of which modules could reference which other modules, and the graph traversal needs to be able to handle all of those cases:

graph of modules referencing other modules

The current implementation relies on module-swapper having been run before tests are run. If not, and the test setup has been migrated over to isolated projects, running tests will fail with a "project_id was not supplied" error.

### DIFF
--- a/infra/blueprint-test/pkg/tft/module_walk_test.go
+++ b/infra/blueprint-test/pkg/tft/module_walk_test.go
@@ -19,6 +19,8 @@ func toSet(args ...string) stringSet {
 func TestFindModulesUnderTest(t *testing.T) {
 	// Monkey patch findReferencedModules() in module_walk.go
 	// to skip parsing real Terraform modules.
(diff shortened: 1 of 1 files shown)


apeabody:
This will go into https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/3189

Issue #3264, opened 2025-11-10
commitlint failures on blueprint-test dependency updates


glasnt:
### TL;DR

The commitlint update in #3143 changes the linting used for conventional commits

Previous PRs used [Conventional Commit GCF](https://github.com/googleapis/repo-automation-bots/blob/main/packages/conventional-commit-lint/README.md)  ([example action result](https://github.com/GoogleCloudPlatform/terraform-dynamic-python-webapp/runs/37830355200))

This tool had previously removed the header-max-length https://github.com/googleapis/repo-automation-bots/pull/4075

The current commitlint tool uses the default rule for header-max-length, which is 100 ([default config](https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/config-conventional/src/index.ts))

This is causing dependency updates to fail for dependency updates like blueprint-test (example https://github.com/GoogleCloudPlatform/terraform-dynamic-python-webapp/pull/390, [example error](https://github.com/GoogleCloudPlatform/terraform-dynamic-python-webapp/actions/runs/19154678289/job/54752938582?pr=390) 117 > 100)

The default rules created by this check may need to be updated to extend or remove this rule. 

### Terraform Configuration

```hcl
GitHub Actions Output
```

### Terraform Version

```sh
GitHub Actions Output
```

PR #3265, opened 2025-11-10
chore: update commitlint and ignore header-max-length

apeabody:
Fix: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/issues/3264


apeabody:
applied

PR #3329, opened 2025-12-29
chore: Add ADC maintainers to GKE Clusters and Cloud DNS

Saranya3:
Add ADC maintainers to https://github.com/terraform-google-modules/terraform-google-cloud-dns, https://github.com/terraform-google-modules/terraform-google-kubernetes-engine


apeabody:
applied

